PR #2, opened 2015-10-28
make rel=noreferrer take priority over referrerpolicy attribute

sc0ttbeardsley:
@annevk @jeisinger as discussed in https://github.com/w3c/webappsec/pull/435 this is an attempt to specify what action is expected when UAs encounter rel=noreferrer and a referrerpolicy attribute on the same element.

The conservative approach is to be backwards-compatible and honor rel=noreferrer.

/cc @wooseok @franziskuskiefer

### DIFF
--- a/index.html
+++ b/index.html
@@ -993,7 +993,7 @@
   <div class="head">
    <p data-fill-with="logo"><a class="logo" href="http://www.w3.org/"> <img alt="W3C" height="48" src="https://www.w3.org/Icons/w3c_home" width="72"> </a> </p>
    <h1 class="p-name no-ref" id="title">Referrer Policy</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2015-10-26">26 October 2015</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2015-10-27">27 October 2015</time></span></h2>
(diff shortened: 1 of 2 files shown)


annevk:
I think this might follow directly from HTML since HTML says that `noreferrer` sets "request's referrer" and that takes precedence over "request's referrer policy".

franziskuskiefer:
agree with @annevk but we should definitely mention it here as there has been confusion about this point.

sc0ttbeardsley:
Great to see this is in line with HTML. Hopefully explicitly stating it here helps someone. Maybe a cross-reference to HTML would help too (I can add a reference just tell me where)? 

BTW, I don't see where HTML explicitly states that the "request's referrer" takes precedence over the "request's referrer policy". I checked here:

https://html.spec.whatwg.org/multipage/semantics.html#link-type-noreferrer
https://html.spec.whatwg.org/multipage/semantics.html#hyperlink-auditing
https://fetch.spec.whatwg.org/#concept-request-referrer
https://fetch.spec.whatwg.org/#concept-referrer-policy

&lt;tangent&gt;
Interestingly there is <a href="https://html.spec.whatwg.org/multipage/semantics.html#link-type-noreferrer">a reference</a> to the Referrer Policy from a section which discusses Ping-From but I don't see any mention of Ping-From in the Referrer Policy, we might want to add something about the Ping-From header somewhere.
&lt;/tangent&gt;

annevk:
Well, the referrer policy spec states which one takes preference. Fetch supplies the request to referrer policy. And HTML invokes Fetch.

jeisinger:
ping-from is tracked here: https://github.com/w3c/webappsec/issues/469

PR #10, opened 2015-11-19
Specify that an empty token is No Referrer

estark37:
There is a use case for making an invalid value return null, so that older browsers ignore newer values that they don't recognize. But there isn't a use case for returning null for an empty referrer policy, so specify that empty should be treated as No Referrer.

This matches what I implemented in https://codereview.chromium.org/1454823004/

@jeisinger PTAL?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1012,7 +1012,7 @@
      <dt>Feedback:
      <dd><span><a href="mailto:public-webappsec@w3.org?subject=%5BREFERRER%5D%20YOUR%20TOPIC%20HERE">public-webappsec@w3.org</a> with subject line “<kbd>[REFERRER] <i data-lt="">… message topic …</i></kbd>” (<a href="http://lists.w3.org/Archives/Public/public-webappsec/" rel="discussion">archives</a>)</span>
      <dt>Issue Tracking:
-     <dd><a href="https://github.com/jeisinger/webappsec-referrer-policy/issues/">GitHub</a>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1012,7 +1012,7 @@ <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="cont
      <dt>Feedback:
      <dd><span><a href="mailto:public-webappsec@w3.org?subject=%5BREFERRER%5D%20YOUR%20TOPIC%20HERE">public-webappsec@w3.org</a> with subject line “<kbd>[REFERRER] <i data-lt="">… message topic …</i></kbd>” (<a href="http://lists.w3.org/Archives/Public/public-webappsec/" rel="discussion">archives</a>)</span>
      <dt>Issue Tracking:
-     <dd><a href="https://github.com/jeisinger/webappsec-referrer-policy/issues/">GitHub</a>
+     <dd><a href="https://github.com/estark37/webappsec-referrer-policy/issues/">GitHub</a>
```
jeisinger: this really should be w3c :-/

@mikewest any idea what to do about this? just ignore?

mikewest: This is coming from Bikeshed's automagical detection of what branch you're on. According to https://github.com/tabatkins/bikeshed/issues/470#issuecomment-140157355, adding `Repository: w3c/webappsec-referrer-policy` to the metadata section at the top of the document will fix this up.

PR #11, opened 2015-11-19
Modify referrer algorithm to use referrerpolicy attribute if present

estark37:
@jeisinger, another one for you to look at, please

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -719,10 +719,9 @@ spec: RFC7231; urlPrefix: https://tools.ietf.org/html/rfc7231
     </li>
     <li>
       If <var>request</var> was initiated by an element with
-      a <code>referrerpolicy</code> attribute, let <var>policy</var> be
-      the result of executing [[#determine-policy-for-token]] on the
-      value of the element's <code>referrerpolicy</code>
-      attribute. Otherwise, let <var>policy</var> be the value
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -713,16 +713,17 @@ <h3 id="determine-requests-referrer">
...
-
   <ol>
     <li>
       Let <var>environment</var> be <var>request</var>'s <code>client</code>.
     </li>
     <li>
-      Let <var>policy</var> be the value of <var>environment</var>'s
-      <a>referrer policy</a>.
+      If <var>request</var> was initiated by an element with
+      a <code>referrerpolicy</code> attribute, let <var>policy</var> be
```
jeisinger: s/attribute/content attribute/

index.src.html
```diff
@@ -713,16 +713,17 @@ <h3 id="determine-requests-referrer">
...
+      the result of executing [[#determine-policy-for-token]] on the
```
jeisinger: the content attribute is already defined such that it can only have valid values, no?

estark37: Ah, right, so this should just be "If request was initiated by an element with a referrerpolicy content attribute, let policy be the attribute's value. Otherwise..."?

jeisinger: sgtm

estark37:
Ok, ptal

PR #16, opened 2016-01-19
Consistency fixes

fmarier:


fmarier:
@estark37 I just rebased this patch, it should apply cleanly now.

fmarier:
I didn't re-build `index.html` because doing so with the latest version of bikeshed touches every other line of the spec :(

estark37:
@fmarier thanks, this looks good. How do we deal with regenerating index.html though? If we don't regenerate it in this commit, won't it just touch every line of the spec the next time someone tries to regenerate it for the next commit? Maybe I should merge this and then regenerate it in a separate commit?

fmarier:
> Maybe I should merge this and then regenerate it in a separate commit?

You could do that, or I could regenerate and add a second commit to this PR. I assume that if we're all using the latest master of bikeshed, the output should be the same?

Let me know if you want me to add that extra commit.

PR #20, opened 2016-02-22
'null' should be set as Origin header if referrer-policy is "no-referrer"

jeisinger:
We take a referrer-policy if "no-referrer" as an indication that the
site does not want to leak its origin in anyway.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1000,7 +1000,7 @@
   <div class="head">
    <p data-fill-with="logo"><a class="logo" href="http://www.w3.org/"> <img alt="W3C" height="48" src="https://www.w3.org/Icons/w3c_home" width="72"> </a> </p>
    <h1 class="p-name no-ref" id="title">Referrer Policy</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2015-12-07">7 December 2015</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2016-02-22">22 February 2016</time></span></h2>
    <div data-fill-with="spec-metadata">
     <dl>
      <dt>This version:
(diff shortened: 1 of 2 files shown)


jeisinger:
@estark37 @annevk 

this updates the spec to match what blink implements.

jeisinger:
@hillbrad who reported this

mikewest:
Looks reasonable, but if we really want `no-referrer` to mean "Dude, don't leak me.", the change you made to `Ping-From` doesn't go far enough, does it?

jeisinger:
yeah, you could argue that "no-referrer" should also omit the "Ping-From" header.

mikewest:
I would argue that. :)

annevk:
Why would you use `ping=""` then though? It seems to me `ping=""` is a fairly explicit signal to leak a URL, similar to a local `referrerpolicy=""` override. Seems weird to require `referrerpolicy=""` whenever you want to use `ping=""`.

mikewest:
That seems like a reasonable way of looking at things, @annevk. That said, I worry a bit about the kinds of use-cases that folks who want to protect things like capability URLs have in mind (e.g. untrusted content). For those kinds of folks, I'd like for there to be a single thing they can set which ensures that the browser doesn't leak the URL. Relying on sanitizers to remove things like the `ping` attribute is tough.

annevk:
Yeah, I guess we could decide that referrer mode is that switch, but we already decided that leaking the origin is fine (CORS), even when referrer mode is none. That means a capability cannot be part of the host. I guess we could restrict `Ping-From` to an origin in certain cases. Would require further patching of HTML.

mikewest:
Sounds like a reasonable compromise.

jeisinger:
@annevk this PR would change that (leaking origin on cors if referrer mode is none)

annevk:
Oh I see, sorry I missed the original intent of this PR, is that really a good idea? I don't think this is what Gecko implemented. @franziskuskiefer?

This would also still leak the origin to foreign fetch service workers I think. Although the details there are still a little underdefined, the way this is fixed is not exactly defense-in-depth.

jeisinger:
This came up in the context of POST requests, where FF just doesn't send an origin header at all...

jeisinger:
(also note that this is what Blink implements - it uses the referrer's origin for the Origin header, not the client's origin)

annevk:
If you specifically mean `POST` requests that result from navigation (e.g., form submission), that is much more reasonable. But this change affects CORS too, which is probably not reasonable and it sounds like Chrome does not implement it for that scenario either.

estark37:
Whether this affects CORS or not, why is it only `no-referrer` and not `no-referrer-when-downgrade` too? That is, if `no-referrer` means "Dude, don't leak me" (as @mikewest succintly put it), wouldn't `no-referrer-when-downgrade` mean "Dude, don't leak me over HTTP"?

franziskuskiefer:
@annevk is right, that's not what gecko implements at the moment. Though I'm not against it as it makes sense not to leak anything in the origin header when someone tries to avoid sending the origin as referer.
But I'm a bit concerned about naming then, `referrerpolicy` doesn't sound like something that's changing the origin header.
@estark37 also made a good point on `no-referrer-when-downgrade`. It would make things more consistent.

jeisinger:
An alternative would be to define the Origin header value to be the referrer header's origin for POST requests - so whenever there's no referrer for whatever reason, we'd also have a null origin.

annevk:
"POST requests" is confusing, since `fetch(..., {method:"POST"})` would not be affected I think. See also https://github.com/whatwg/xhr/issues/31 by the way. @tyoshino and @hiroshige-g have been trying to figure out the correct architectural setup for the `Origin` header. It's multipurposeness has not done it much good.

jeisinger:
form submissions then?

annevk:
Yeah, that's what I think you mean.

Issue #15, opened 2015-12-17
Allow referrerpolicy attribute on <link>


igrigorik:
If I'm reading current draft correctly, referrerpolicy is restricted to a, area, img, and iframe? Can we extend it to `<link>` as well? This would enable developers to control the policy of prefetch/preload/prerender resources (also, stylesheets and other types fetched via `link`). 

```html
<link rel=prefetch as=image href=example.com/thing.jpg referrerpolicy=origin>
```

/cc @mikewest

estark37:
That seems reasonable to me, what do you think @jeisinger?

davidillsley:
Similarly, is there a reason why its not available on ```<script>```? (I don't see why a CDN hosting JS should necessarily see the referer, even if some might want at least the origin for analytics)

jeisinger:
The intent of the referrerPolicy attribute is to allow for setting a different policy for outgoing navigations.

If you need a finer control over the referrer for sub resources, please use a service worker and modify the referrer via the fetch API.

mikewest:
> The intent of the referrerPolicy attribute is to allow for setting a different policy for outgoing navigations.

I thought ads were also a use case y'all wanted to support, hence `<img referrerpolicy=whatever>` inside a page that set a policy of `none`. It doesn't seem like you lose much by widening that implementation to other types of requests, given that it's already being piped through for images.

jeisinger:
Dunno, I don't really like the idea of adding this to every element that can load something from the network :-/

davidillsley:
Does the delivery by CSP apply to just a, area, img or iframe elements?

jeisinger:
Delivery via CSP was removed from the spec, and replaced by a dedicated header. That header applies to everything.

jeisinger:
Dominic came and explained to me the use case for prefetch / prerender. I can see how that's more like an outgoing navigation, and won't be covered by my service worker proposal.

I guess even if we didn't want to add the attribute to all resources, we could still spec it for prefetch/prerender.

annevk:
Just to be clear, we already allow it for resources other than navigation. `<img>` was pointed out, `fetch()` is another one. I think it would make sense to support it for other elements too, ever since `rel=noreferrer` existed folks have been asking for that. Why don't you like that idea?

jeisinger:
Mainly that it blows up the spec so much.

annevk:
Maybe now is the time then to move the attribute definition to HTML. 😛

[...]

PR #22, opened 2016-03-08
Add referrerpolicy attribute on link elements

estark37:
Addresses issue #15 

@jeisinger PTAL?

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -134,6 +134,10 @@ spec: RFC2616; urlPrefix: https://tools.ietf.org/html/rfc7231
     text: redirection 3xx; url: section-6.4
 </pre>
 
+<pre class="link-defaults">
+spec:html; type:element; text:link
+</pre>
+
 <!--
 ████ ██    ██ ████████ ████████   ███████
  ██  ███   ██    ██    ██     ██ ██     ██
@@ -380,7 +384,7 @@ spec: RFC2616; urlPrefix: https://tools.ietf.org/html/rfc7231
     </li>
     <li>
       Via a <code>referrerpolicy</code> content attribute on an <{a}>,
-      <{area}>, <{img}> or <{iframe}> element.
+      <{area}>, <{img}>, <{iframe}>, or <{link}> element.
     </li>
     <li>
       Implicitly, via inheritance.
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -650,6 +655,17 @@ <h4 id="html-iframe-element-attributes">Attributes</h4>
...
+  <h3 id="html-link-element">HTMLLinkElement</h3>
+  <pre class="idl">
+    partial interface HTMLLinkElement {
+      attribute DOMString referrerPolicy;
+    };
+  </pre>
+  <h4 id="html-link-element-attributes">Attributes</h4>
+  The <code>referrerPolicy</code> IDL attribute must <a>reflect</a> the
+  <code>referrerpolicy</code> content attribute, <a>limited to only known
+  values</a>.
```
jeisinger: I wonder whether we need to say something about how the referrerPolicy attribute is supposed to be applied here, or whether that's ok to be left open for interpretation?

estark37: Hmm... we don't say anything for the other elements, do we? I feel like it's pretty obvious how the attribute is applied, unless there's some `<link>` nuance that I don't know about (which would not be surprising).

annevk: Ideally it's defined in detail. E.g., how it's passed to Fetch and such. That's definitely part of the requirement to upstream this to HTML.

jeisinger: well, fetch doesn't talk about how it fetches css, or favicons, or what happens in the bookmark system, or how prerendering would work...

annevk: HTML defines how favicons and CSS are fetched. CSS should define how its subresources are fetched. Bookmarks navigate, which HTML defines and use fetch. I don't know what specification defines prerendering.

jeisinger: would you expect a bookmark to remember the referrer for its future navigations?

anyway, I guess the best we can do is saying something like if this link elements triggers a fetch or a navigation that creates a referrer from the current document, the referrer policy should apply.

annevk: I would expect bookmarks to always navigate without a referrer, but that's mostly up to the user agent since it's a UI feature.

estark37: Sorry for the delay. I still don't really get what's special about the link element that it needs special treatment here. The integration with Fetch is already defined by using Section 8.3 "Determine request's Referrer", which includes "If request was initiated by an element with a referrerpolicy content attribute..."

jeisinger: yeah, this PR LGTM

If there are issues with our general fetch integration, let's address them separately

foolip:
Looks like https://w3c.github.io/webappsec-referrer-policy/ isn't automatically in sync with master? Any guess about when it'll include the `HTMLLinkElement` extensions?

jeisinger:
no, we have to manually push to gh-pages... let me quickly do that

foolip:
Thanks! https://w3c.github.io/webappsec-referrer-policy/#html-link-element

Issue #5, opened 2015-11-11
Define how CSS interacts with spec for the sake of defining how referrers should be handled


jeisinger:
Esp for resources loaded from stylesheets, and stylesheets loaded from stylesheets.

@mikewest @estark37 @annevk @tabatkins

annevk:
@bzbarsky if you have a moment I'd appreciate if you could take a look at this. The question is what to do with referrer and stylesheets. I believe we talked about it before and you felt pretty strongly that the referrer being the stylesheet's URL for any subresources of the stylesheet was the way to go (and what Gecko implements). Presumably if a stylesheet references another stylesheet, the subresources of that second stylesheet would use that second stylesheet's URL.

https://github.com/w3c/webappsec-referrer-policy/pull/68#pullrequestreview-1198995 is probably also worthwhile to read as it sketches some other scenarios.

bzbarsky:
Talked about this with folks today.  The conclusion seems to be to use the stylesheet URI as the referrer and use the sheet's referrer policy (which may be set via headers).  Need tests, especially for the "set via headers" thing.

domenic:
Should we do the same thing for module scripts? I'm particularly talking about import statements; I assume fetch() calls and such use that of their relevant settings object.

domenic:
Oh, and does the referrerpolicy attribute on the link element impact anything beyond the initial fetch? For module scripts, a few attributes currently carry over (crossorigin and nonce IIRC).

bzbarsky:
For CSS right now `crossorigin` does not carry over to anything beyond the initial fetch.  I don't think `referrerpolicy` should either, personally.

domenic:
I'm certainly willing to change script to align with that for crossorigin and not carry over beyond the initial fetch.

And I guess also change scripts so that they can get a referrer policy from the header.

But what about nonce? @mikewest pretty explicitly made that carry over into script imports. Is nonce different? Should it apply to stylesheet imports?

jeisinger:
I don't think anybody is proposing to put an referrer policy attribute on <link rel=stylesheet> here.

For the other attributes, I'd propose to move the discussion over to the respective specs.

domenic:
link already has a referrerpolicy attribute.

I'm hoping we can at some point hash out a consistent story for all the fetch-options attributes.

annevk:
I think for module scripts copying crossorigin is needed. Otherwise you get opague module scripts.

domenic:
[...]

PR #34, opened 2016-04-21
Add a paragraph talking about how to integrate with CSS

jeisinger:
@annevk @estark37 @tabatkins 

ptal

### DIFF
--- a/index.html
+++ b/index.html
@@ -1000,7 +1000,7 @@
   <div class="head">
    <p data-fill-with="logo"><a class="logo" href="http://www.w3.org/"> <img alt="W3C" height="48" src="https://www.w3.org/Icons/w3c_home" width="72"> </a> </p>
    <h1 class="p-name no-ref" id="title">Referrer Policy</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2016-03-28">28 March 2016</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2016-04-21">21 April 2016</time></span></h2>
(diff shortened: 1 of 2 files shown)


jeisinger:
addresses #5

tabatkins:
You probably have a good reason, but why does this specify a different policy than what CSS Fonts specifies?  (Fonts says the referrer should be the stylesheets url, you're saying it should be the containing document's url.)  What's the pro/con of doing it the way you specify vs what Fonts says?

jeisinger:
Pro of mine is: does what all other requests do. CSS fonts also sets the
origin to the containing document. The stylesheet has no referrer policy,
but the doc has one

Pro of CSS fonts: none afaik. I asked the person who wrote the blink
implementation, and filed a spec bug, and asked you, but nobody knew any
particular reason.

tabatkins:
Sounds reasonable, especially the fact that stylesheets don't have a referrer policy.

Can you see if impls are willing to change @font-face loading to harmonize with this? We can change the spec easily then.

annevk:
The referrer for CSS subresources should be that of the stylesheet. Just like the base URL is the URL of the stylesheet. I think most implementations have behaved this way ever since CSS was invented.

annevk:
(This should be defined in CSS though, ideally...)

jeisinger:
meh

in the end, I guess that means that stylesheets should track their own referrer policy, and we can probably only deliver it via the referrer header.

annevk:
That makes sense, I guess we could say that they inherit one from the document. All this would be easier to define if CSS had some infrastructure around its basic model.

jeisinger:
hum, so if you load it via <link rel=stylesheet referrerpolicy=origin> but the document's referrerpolicy is never, which one should be inherited?

jeisinger:
in other words, I'd rather not inherit from the document.

annevk:
I can't quite parse that sentence. The suggestion was to inherit if no policy was specified through HTTP.

PR #35, opened 2016-04-21
Clarify that modifying img.referrerPolicy is not a relevant mutation

### LINKED ISSUE #33: Should the referrerPolicy attribute be part of the "relevant mutations" for HTMLImageElement?
https://html.spec.whatwg.org/multipage/embedded-content.html#relevant-mutations defines a set of mutations that cause the image element to request the image data. There's no text in this specification that indicates how modifications to the `referrerPolicy` attribute are supposed to affect `` after a load has previously been started (whether it has previously completed or not).

jeisinger:
@estark37 @annevk 

fixes #33

### DIFF
--- a/index.html
+++ b/index.html
@@ -1425,6 +1425,7 @@
     <h4 class="heading settled" data-level="5.3.1" id="html-image-element-attributes"><span class="secno">5.3.1. </span><span class="content">Attributes</span><a class="self-link" href="#html-image-element-attributes"></a></h4>
      The <code>referrerPolicy</code> IDL attribute must <a data-link-type="dfn" href="http://www.w3.org/TR/html5/infrastructure.html#reflect">reflect</a> the <code>referrerpolicy</code> content attribute, <a data-link-type="dfn" href="http://www.w3.org/TR/html5/infrastructure.html#limited-to-only-known-values">limited to only known
   values</a>. 
+    <p class="note" role="note">Note: Modifying the <code>referrerPolicy</code> IDL attribute is not a <a data-link-type="dfn" href="http://www.w3.org/TR/html5/embedded-content-0.html#relevant-mutations">relevant mutation</a> for an <code><a data-link-type="element" href="https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element">img</a></code> element.</p>
     <h3 class="heading settled" data-level="5.4" id="html-iframe-element"><span class="secno">5.4. </span><span class="content">HTMLIFrameElement</span><a class="self-link" href="#html-iframe-element"></a></h3>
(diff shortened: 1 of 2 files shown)


annevk:
LGTM

annevk:
Per @tungmangtdh3 suggested change to web-platform-tests I now wonder if maybe it would be better if this did trigger a rerun. @zcorpan, is `` really that synchronous? It might be surprising that
```
img.referrerPolicy = y
img.src = x
```
and
```
img.src = x
img.referrerPolicy = y
```
have different results.

tungmangtdh3:
I tried with Firefox nightly
img.referrerPolicy = y
img.src = x
Pass.
img.src = x
img.referrerPolicy = y
Failed.
I have not tried to compare result of 2 cases with Google Canary and Opera.

jeisinger:
I agree that this might be confusing, but once the request hit the network, modifying the policy is too late, so just resending the request won't unleak the referrer.

annevk:
Does `` end up creating two requests for the above scenarios for attributes that are relevant mutations?

jeisinger:
hum, setting e.g. crossOrigin to "anonymous" doesn't trigger a second request.

jeisinger:
@zcorpan can you advise what the right thing to do here is?

zcorpan:
Per spec the "update the image data" algorithm starts synchronously when a relevant mutation happens, but it only changes the image synchronously if the `img` has only `src` (no `srcset` or `picture` parent) *and* that URL is in the image cache. Otherwise, it "awaits a stable state" (microtask) to continue the algorithm, which lets the current script complete and set all attributes it wants before the actual request is created. So the order in which a script sets src/crossOrigin/referrerPolicy should not matter per spec.

zcorpan:
cc @yoavweiss

jeisinger:
ptal

zcorpan:
You need to change `index.src.html` (and regen).

I don't have an opinion yet on whether this should be a relevant mutation or not. It does not make any difference for setting things up initially in a script. If it is a relevant mutation, then changing `referrerPolicy` *later* will cause "update the image data" to re-run.

That algorithm doesn't currently know about `referrerPolicy` (e.g. it doesn't use it as part of the key in the image cache). Should it?

zcorpan:
I'd prefer "must" over "should" and also clarify what modifications means; in particular setting it to the same value as the previous value

jeisinger:
I guess in my test case, I had the image already in the cache then.

I don't have a strong opinion either. Not making it a relevant modification is easier, as this won't require implementation changes :)

annevk:
If it only has an affect when changing `referrerPolicy` later than we should probably not do it. The only reason I changed my mind was https://github.com/w3c/webappsec-referrer-policy/pull/35#issuecomment-213271557 but if that's not affected I think we're okay.

jeisinger:
ok, and back to *not* a relevant mutation...

zcorpan:
LGTM :-)

yoavweiss:
belated LGTM :)

PR #36, opened 2016-04-21
Rewrite Fetch integration, especially around redirects

estark37:
This cleans up the Fetch integration in the following ways:

- Integrate the handling of Referrer-Policy headers on redirects with
  the HTTP-redirect fetch algorithm. This (I think) removes the need for
  responses to have associated referrer policies, and is also more
  precise/clean.

- Insert a TODO to define an HTML integration for the referrerpolicy
  content attributes. This removes the need for handwavy handling of the
  attribute when determining the Referer for a request.

- Use the request's referrer policy when determining the Referer for a
  request (no need to consult the environment's policy because that is
  now accounted for in Fetch as of
  https://github.com/whatwg/fetch/pull/280). Issue 26

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -24,15 +24,22 @@ spec: DOM; urlPrefix: http://www.w3.org/TR/dom/
 spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
     text: fetching
+    text: request; url: concept-request
     text: response; url: concept-response
   type: interface
     text: Request
+  type: interface
+    text: Response
   type: attribute
     text: url; for: Request; url: concept-request-url
     text: origin; for: Request; url: concept-request-origin
     text: client; for: Request; url: concept-request-client
     text: context; for: Request; url: concept-request-context
     text: context-frame-type; for: Request; url: concept-request-context-frame-type
+spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
+  type: dfn
+    urlPrefix: browsers.html
+      text: opaque origin; url: concept-origin-opaque
 spec: HTML5; urlPrefix: http://www.w3.org/TR/html5/
   type: element-attr
     urlPrefix: document-metadata.html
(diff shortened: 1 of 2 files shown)


estark37:
@jeisinger @annevk @yutakahirano wdyt?

annevk:
Okay, so the idea is that for redirects we update request's referrer policy based on the response's header list, regardless of what type of request this is.

And then, if this request is for a document or worker, we set their "environment" referrer policy to that given by the response's header list too.

That seems pretty good.

index.src.html
```diff
@@ -768,87 +790,66 @@ <h3 id="determine-requests-referrer">
...
-
-              <ol>
-                <li>
-                  Let <var>document</var> be the {{Document}} object of the
-                  <a>active document</a> of the <a>browsing context</a> of
-                  <var>environment</var>'s <a>responsible browsing context</a>.
-                </li>
-              </ol>
+              If <var>document</var>'s <a>origin</a> is not a scheme/host/port
+              tuple (because, for example, it has been sandboxed into a unique
```
annevk: You can use "opaque origin" here rather than negating scheme/host/port. See https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque.

index.src.html
```diff
@@ -752,13 +757,30 @@ <h3 id="set-referrer-policy">
     </li>
   </ol>
 
+  <h3 id="set-requests-referrer-policy-on-redirect">
+    Set |request|'s referrer policy on redirect
+  </h3>
+
+  Given a {{Request}} |request| and a {{Response}} |actualResponse|,
```
yutakahirano: This `Request` should be [[concept-request]](https://fetch.spec.whatwg.org/#concept-request). Ditto for Response.

yutakahirano:
lgtm, thank you!

index.src.html
estark37: Done, thanks.

index.src.html
estark37: Done, I think (did I do that right?). Also, presumably this should apply to all the other instances of this in various algorithms also? If so I'll do that separately.

estark37:
@jeisinger do you want to take a look before I merge this?

index.src.html
annevk: Yeah, Fetch never JavaScript objects to Referrer Policy, just concepts.

jeisinger:
LGTM, merge it!

PR #41, opened 2016-04-28
Clarify that controlling the referrer isn't enough to stop capability URLs to leak

### LINKED ISSUE #38: Clarify that the referrer policy does not affect CSP reports...
... maybe

jeisinger:
closes #38 

/cc @estark37


estark37:
lgtm

PR #42, opened 2016-04-29
Integrate with HTML, part 1 of n

domenic:
This is the counterpart to https://github.com/whatwg/html/pull/1153, which removes the parts of the spec that are now defined in HTML: the content and IDL attributes on various elements, plus the `<meta http-equiv>` extension.

There is still much to do; I will open a separate issue detailing the plan.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,55 +1,319 @@
 <!doctype html><html lang="en">
  <head>
   <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
+  <meta content="width=device-width, initial-scale=1, shrink-to-fit=no" name="viewport">
   <title>Referrer Policy</title>
-<style data-fill-with="stylesheet">/* This section copied from the W3C's ED styles. */
+<style data-fill-with="stylesheet">/******************************************************************************
+ *                   Style sheet for the W3C specifications                   *
+ *
+ * Special classes handled by this style sheet include:
+ *
+ * Indices
+ *   - .toc for the Table of Contents (<ol class="toc">)
+ *     + <span class="secno"> for the section numbers
+ *   - #toc for the Table of Contents (<nav id="toc">)
+ *   - ul.index for Indices (<a href="#ref">term</a><span>, in §N.M</span>)
+ *   - table.index for Index Tables (e.g. for properties or elements)
+ *
+ * Structural Markup
+ *   - table.data for general data tables
+ *     -> use 'scope' attribute, <colgroup>, <thead>, and <tbody> for best results !
+ *     -> use <table class='complex data'> for extra-complex tables
+ *     -> use <td class='long'> for paragraph-length cell content
+ *     -> use <td class='pre'> when manual line breaks/indentation would help readability
+ *   - dl.switch for switch statements
+ *   - ol.algorithm for algorithms (helps to visualize nesting)
+ *   - .figure and .caption (HTML4) and figure and figcaption (HTML5)
+ *     -> .sidefigure for right-floated figures
+ *   - ins/del
+ *
+ * Code
+ *   - pre and code
+ *
+ * Special Sections
+ *   - .note       for informative notes             (div, p, span, aside, details)
+ *   - .example    for informative examples          (div, p, pre, span)
+ *   - .issue      for issues                        (div, p, span)
+ *   - .advisement for loud normative statements     (div, p, strong)
+ *   - .annoying-warning for spec obsoletion notices (div, aside, details)
+ *
+ * Definition Boxes
+ *   - pre.def   for WebIDL definitions
+ *   - table.def for tables that define other entities (e.g. CSS properties)
+ *   - dl.def    for definition lists that define other entitles (e.g. HTML elements)
+ *
+ * Numbering
+ *   - .secno for section numbers in .toc and headings (<span class='secno'>3.2</span>)
+ *   - .marker for source-inserted example/figure/issue numbers (<span class='marker'>Issue 4</span>)
+ *   - ::before styled for CSS-generated issue/example/figure numbers:
+ *     -> Documents wishing to use this only need to add
+ *        figcaption::before,
+ *        .caption::before { content: "Figure "  counter(figure);  }
+ *        .example::before { content: "Example " counter(example); }
+ *        .issue::before   { content: "Issue "   counter(issue);   }
+ *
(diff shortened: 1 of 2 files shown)


jeisinger:
Can you please update the "Integration with HTML section" to point out the changes that need to be done to the W3C HTML spec (and preferably file spec bugs for them), similar to what we have in CSP: https://w3c.github.io/webappsec-csp/#html-integration

domenic:
Can do, although I don't think that's really necessary...

jeisinger:
for better or worse, this spec should make sense in the W3C context as well

domenic:
Done

Ms2ger:
I disagree that there's any point to mentioning the harmful w3c fork.

jeisinger:
@Ms2ger please bring this up with Wendy, who's working on this issue

mikewest:
Two broad points:

1. I think it is good for a document describing a single feature to describe all of it in a way that web developers and browser implementers can skim in one sitting, rather than hopping back and forth between HTML, DOM, Fetch, and etc. Those are and should be the normative references, but this document should have enough on its bones to be a cohesive picture of referrer policy. Dropping the integration text entirely seems counterproductive. I think CSP has a reasonable compromise at the moment, and the effort required to maintain that section is IMO outweighed by its value to the reader.

2. I agree that the split between whatwg/html and w3c/html is frustrating, annoying, duplicitous, error-prone, and likely to result in developer confusion. As long as we're in a world in which folks are working on both, I think it's reasonable for us to try to upstream to both. That said, I personally am not spending time either reviewing or creating PRs for w3c/html, as it's not the document I use when implementing. I do file bugs against that spec, and I think it's worth doing so, if only to make clear the amount of effort involved in maintaining and updating the document.

One narrower point: as long as we're driving this spec through the WebAppSec WG, the snapshots we produce will need to reference documents that body considers stable. It's totally fine for the ED to reference WHATWG docs; @wseltzer, @dveditz, and @hillbrad are going to need to work out how to get those documents past WD to REC.

For better or worse, this is a problem that we've been kicking down the field starting with the CSP2 last call ~2 years ago. We've got a face-to-face meeting in Mountain View on the 16th/17th. I hope we can make some progress there.

domenic:
Can you expand on what your (1) means in practice for this PR? Are you saying that the content and IDL attributes should not move to HTML?

mikewest:
Naah. They should absolutely move to HTML. For this document, I'd suggest that we mention them, explain their purpose, and etc, in a non-normative "Integration with HTML" section, similar conceptually to what the CSP draft is doing with various HTML concepts.

I'd like folks to be able to read _this_ document, and understand everything they need to understand. Non-normative duplication that points out to an official source seems like the right way to handle that.

domenic:
That makes sense. I'd ideally like to leave that to the authors of this document, since they're in a better position to do so; I included a couple sentences as a pointer, which was about as much more information as existed in the spec previously, but I think you are imaging more of a lengthy tutorial.

I am still hoping this can be merged soon-ish so we can continue with the HTML integration plan described in #43.

domenic:
Although I guess this should not be merged until https://github.com/whatwg/html/pull/1153 is reviewed.

estark37:
This PR lgtm. Once it's merged I can take a look through and see if we need a separate HTML integration section.

estark37:
Forgot to say, thanks for doing this work @domenic! And sorry I'm sooooooo slow reviewing.

tabatkins:
> One narrower point: as long as we're driving this spec through the WebAppSec WG, the snapshots we produce will need to reference documents that body considers stable. It's totally fine for the ED to reference WHATWG docs; @wseltzer, @dveditz, and @hillbrad are going to need to work out how to get those documents past WD to REC

There is absolutely nothing wrong with reffing WHATWG specs in later W3C stability levels. You just need to show that they are reasonably stable, like reffing any other document outside of W3C.

mikewest:
@tabatkins: I agree with you, so I'll rephrase for clarity: so far I have not been successful when attempting to documents that reference WHATWG's HTML or Fetch to CR+.

domenic:
OK, @estark37 I updated this to also include the switch to string-named referrer policies. It seems like a really nice improvement, upon viewing the results. The document was rather inconsistent in how it referred to the policies, and this totally clarifies the empty string business.

Would love to get this merged!

estark37:
Thanks! I'll go ahead and merge this but @jeisinger I think it would still be good for you take a look when you have a chance.

domenic:
\o/! I'll get HTML updated posthaste.

domenic:
Oh, how do we update gh-pages? It looks like it is behind master by a decent bit, not just these commits.

estark37:
@domenic yeah, I noticed that the other day... I'm not sure how we do that. I suppose I'll just trying resetting it to master and push -f and see what happens. Wish me luck

domenic:
I think `make publish` maybe? But on the computer of someone with access to the repo.

estark37:
Oh, cool, that did it. I guess I have access to the repo. :)

jdm:
Can we update the referrer-policy tests in https://github.com/w3c/web-platform-tests/ as well?

domenic:
What is there to update as a result of this PR?

jdm:
The change from `origin-only` to `origin`?

domenic:
The string did not change, only the spec section title





On Tue, May 24, 2016 at 1:56 PM -0700, "Josh Matthews" <notifications@github.com<mailto:notifications@github.com>> wrote:


The change from origin-only to origin?

-

jdm:
@domenic ae3e8933125caa5f39eeadff235eeee49ac12ee1 appears to disagree with you? Or maybe a16588962af5c59e8c056fb852f98293c00b5862 is more relevant, but it looks like the string was changed without updating the tests.

domenic:
The string was always "origin". The state was name "Origin Only", which caused confusion. When I removed the state name vs. string distinction, I accidentally made a normative change to change the string from "origin" to "origin-only". This was accidental, so I reverted it in the commits you cite.

domenic:
You can see the old spec at https://rawgit.com/w3c/webappsec-referrer-policy/5874e73ddf780ca90a4e0b206df3194c393648e3/index.html to confirm that the string has always been "origin".

jdm:
I see. Thanks for the clarification!

PR #49, opened 2016-05-17
Integrate with HTML, part 2 of n

### LINKED ISSUE #30: Please work with the HTML spec to not monkeypatch browsing context creation
https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-delivery-implicit-nested is inserting steps into browsing context creation that are not at all obvious while reading the HTML spec that defines creation of browsing contexts.  This should probably be integrated better with HTML (e.g. via an explicit hook HTML invokes).

@annevk, thoughts?

### LINKED ISSUE #32: Verbiage about creating nested browsing contexts for particular URLs does not make sense
When nested browsing contexts are created the URL is not known (all browsing contexts are created with an about:blank document in them), and in fact can change over the lifetime of the browsing context.

For example, you can have an `<iframe srcdoc>` that then has the `srcdoc` attribute removed and hence loads its `src`.  Or you can have an iframe that loads some http page and then later loads a blob URL.  This does NOT create a new nested browsing context.

What you probably need is to do is do the current inheritance unconditionally on nested browsing context creation (so that initial about:blank inherits the referrer policy, but note that there's the open question of whether it should inherit a snapshot or an alias) and then do things for srcdoc and in fetch that inherit referrer policy the way origin is inherited... possibly including the aliasing origin does.

domenic:
This is the counterpart to https://github.com/whatwg/html/pull/1205, reflecting the Referrer Policy spec to no longer try to specify "implicit delivery". It also creates a dedicated section for the empty-string referrer policy.

This closes #30 and closes #32.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -54,7 +54,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: browsing context container
       text: child browsing context
       text: creating a new Document object
-      text: navigated
+      text: navigation; url: navigate
       text: nested browsing context
       text: nested through; url: browsing-context-nested-through
       text: opener browsing context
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -514,24 +504,29 @@ <h2 id="integration-with-fetch">Integration with Fetch</h2>
...
   Issue: TODO: define content attribute integrations. For example, for
   img elements, HTML should set the request's associated referrer policy
   before fetching.
 
   Note: W3C HTML5 does not define the <code>referrerpolicy</code> content
   attributes, or <code>referrerPolicy</code> IDL attributes, or the
-  <a for="meta"><code>referrer</code></a> keyword for <{meta}>. For this spec to
-  make sense with W3C HTML5, those would need to be copied from [[HTML]].
+  <a for="meta"><code>referrer</code></a> keyword for <{meta}>, or the
+  integration with navigation or running a working. For this spec to make sense
```
estark37: working => worker

domenic: Thanks, fixed and pushed new commit.

estark37:
Thanks @domenic. @annevk would you have time to take a look at this as well?

domenic:
(Also, I rebased on master, so there shouldn't be any more conflicts.)

annevk:
Bikeshed text looks okay. Didn't look at the generated output.

Issue #50, opened 2016-05-17
Add the non-downgrading versions of origin and origin-when-cross-origin


fmarier:
As discussed [on the list](https://lists.w3.org/Archives/Public/public-webappsec/2016Mar/0085.html), in the [last teleconf](https://www.w3.org/2016/04/20-webappsec-minutes.html#item09) and in the Face-to-face, I will follow up #19 with spec changes for the other two proposed policy states.

fmarier:
As [mentioned](https://github.com/w3c/webappsec-referrer-policy/pull/19#issuecomment-220077209) in #19, once these new states are in, both Fetch and HTML will need to be updated too.

fmarier:
And the [Security Considerations](https://github.com/w3c/webappsec-referrer-policy/blob/0f9d26e8702faf8130f62fdc7afc004ce2ee7a16/index.src.html#L900-L923) should probably mention the new states too.

fmarier:
The [web platform tests](https://github.com/w3c/web-platform-tests/tree/master/referrer-policy) will also need to be updated for the three new states (the third one being `same-origin` which has already been merged into the spec).

fmarier:
Reopening since I still need to update:

- Fetch
- HTML
- Web Platform tests

fmarier:
HTML changes here: https://github.com/whatwg/html/pull/1589

annevk:
(Updated your comment to point to the Fetch and WPT issues.)

fmarier:
The tests are done.

PR #19, opened 2016-01-19
Introduce a "Same-Origin Only" policy state

fmarier:
Except for "No Referrer", the existing policy states do not allow authors to use a policy that maintains the downgrade protection of the default referrer policy and also limit the amount of information that is leaked via the `Referer` header to just the origin.
    
This new policy state allows an author to maintain full referrer information for their internal pages while not leaking anything out to third-parties.

(I am planning on updating the privacy considerations section in a future commit.)

PR #52, opened 2016-05-31
Add the policy states safe-origin and safe-origin-when-cross-origin.

mozfreddyb:
This is a follow-up of #19. It was already discussed in #50.

The web platform tests still need to be updated for the three new states (the third one being same-origin which has already been merged into the spec). We also want this upstreamed in Fetch and HTML.

Flagging @estark37 for review. @domenic and @annevk for HTML and Fetch respectively.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -505,7 +505,7 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
     "Referrer-Policy:" 1#<a>policy-token</a>
   </pre>
   <pre>
-    <dfn export>policy-token</dfn>   = "no-referrer" / "no-referrer-when-downgrade" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
+    <dfn export>policy-token</dfn>   = "no-referrer" / "no-referrer-when-downgrade" / "safe-origin" / "safe-origin-when-cross-origin" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
   </pre>
 
   Note: The header name does not share the HTTP Referer header's misspelling.
@@ -757,6 +757,53 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
         <dt><a>"<code>unsafe-url</code>"</a></dt>
         <dd>Return <var>referrerURL</var>.</dd>
 
+        <dt><a>"<code>safe-origin</code>"</a></dt>
+        <dd>
+          <ol>
+            <li>
+              If |environment| is not null:
+              <ol>
+                <li>
+                  If |environment| is <a>TLS-protected</a> <em>and</em>
+                  the <a>origin</a>
+                  of <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-current-url">current
+                  URL</a> is not an <a><em>a priori</em> authenticated
+                  URL</a>, then return <code>no referrer</code>.
+                </li>
+              </ol>
+            </li>
+            <li>Return |referrerOrigin|.
+          </ol>
+        </dd>
+
+        <dt><a>"<code>safe-origin-when-cross-origin</code>"</a></dt>
+        <dt>the empty string</dt>
+        <dd>
+          <ol>
+            <li>
+              If |environment| is not null:
+              <ol>
+                <li>
+                  If |environment| is <a>TLS-protected</a> <em>and</em>
+                  the <a>origin</a>
+                  of <var>request</var>'s <a href="https://fetch.spec.whatwg.org/#concept-request-current-url">current
+                  URL</a> is not an <a><em>a priori</em> authenticated URL</a>
+                  <ol>
+                    <li>
+                      If <var>request</var> is a <a>same-origin request</a>, then
+                      return <var>referrerURL</var>.
+                    </li>
+                    <li>
+                      Otherwise, return <var>referrerOrigin</var>.
+                    </li>
+                  </ol>
+                </li>
+              </ol>
+            </li>
+            <li>Return <code>no referrer</code>.
+          </ol>
+        </dd>
+
         <dt><a>"<code>same-origin</code>"</a></dt>
         <dd>
           <ol>
@@ -878,6 +925,15 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
     </li>
     <li>
       If <var>token</var> is an <a>ASCII case-insensitive match</a> for the
+      string "<code>safe-origin</code>", return <a>"<code>safe-origin</code>"</a>.
+    </li>
+    <li>
+      If <var>token</var> is an <a>ASCII case-insensitive match</a> for the
+      string "<code>safe-origin-when-cross-origin</code>", return
+      <a>"<code>safe-origin-when-cross-origin</code>"</a>.
+    </li>
+    <li>
+      If <var>token</var> is an <a>ASCII case-insensitive match</a> for the
       string "<code>same-origin</code>", return <a>"<code>same-origin</code>"</a>.
     </li>
     <li>
@@ -932,7 +988,7 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
 
   Authors can use the policy states <a>"<code>safe-origin</code>"</a>,
   <a>"<code>safe-origin-when-cross-origin</code>"</a> and
-  <a>"<code>no-referrer</code>"</a> to preserve sensitive information from
+  <a>"<code>no-referrer</code>"</a> to prevent sensitive information from
   leaking via referrer headers. [[CAPABILITY-URLS]]
 
   <h3 id="downgrade">Downgrade to less strict policies</h3>
--- a/index.src.html
+++ b/index.src.html
@@ -927,8 +927,13 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
   <a>"<code>unsafe-url</code>"</a> might leak the origin and the URL of
   a secure site respectively via insecure transport.
 
-  Those two policies are include in the spec nevertheless to lower the friction
+  Those two policies are included in the spec nevertheless to lower the friction
   of sites adopting secure transport.
+
+  Authors can use the policy states <a>"<code>safe-origin</code>"</a>,
+  <a>"<code>safe-origin-when-cross-origin</code>"</a> and
+  <a>"<code>no-referrer</code>"</a> to preserve sensitive information from
+  leaking via referrer headers. [[CAPABILITY-URLS]]
 
   <h3 id="downgrade">Downgrade to less strict policies</h3>
 
--- a/index.src.html
+++ b/index.src.html
@@ -332,7 +332,7 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
     priori</em> authenticated URLs</a>.
   </div>
 
-  <h3 dfn export id="referrer-policy-safe-origin" oldids="referrer-policy-state-safe-origin">"<code>safe-origin</code></h3>
+  <h3 dfn export id="referrer-policy-safe-origin" oldids="referrer-policy-state-safe-origin">"<code>safe-origin</code>"</h3>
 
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -869,9 +927,14 @@ <h3 id="information-leakage">Information Leakage</h3>
   <a>"<code>unsafe-url</code>"</a> might leak the origin and the URL of
   a secure site respectively via insecure transport.
 
-  Those two policies are include in the spec nevertheless to lower the friction
+  Those two policies are included in the spec nevertheless to lower the friction
   of sites adopting secure transport.
 
+  Authors can use the policy states <a>"<code>safe-origin</code>"</a>,
+  <a>"<code>safe-origin-when-cross-origin</code>"</a> and
+  <a>"<code>no-referrer</code>"</a> to preserve sensitive information from
```
fmarier: typo: s/preserve/prevent/

fmarier:
Thanks for putting that together @mozfreddyb. There  more three places where the new states need to be added:

- https://github.com/w3c/webappsec-referrer-policy/blob/df68c16003b7f89bbaafe44d8756720889f9d64a/index.src.html#L450
- https://github.com/w3c/webappsec-referrer-policy/blob/df68c16003b7f89bbaafe44d8756720889f9d64a/index.src.html#L690
- https://github.com/w3c/webappsec-referrer-policy/blob/df68c16003b7f89bbaafe44d8756720889f9d64a/index.src.html#L806-L841

index.src.html
```diff
@@ -332,6 +332,32 @@ <h3 dfn export id="referrer-policy-origin" oldids="referrer-policy-state-origin"
     priori</em> authenticated URLs</a>.
   </div>
 
+  <h3 dfn export id="referrer-policy-safe-origin" oldids="referrer-policy-state-safe-origin">"<code>safe-origin</code></h3>
```
fmarier: This is missing a `"` (double quote) between `</code>` and `</h3>`.

Without that the links to that section don't work and bikeshed issues an error:

    LINK ERROR: No 'dfn' refs found for '"safe-origin"'.

index.src.html
```diff
@@ -699,6 +757,53 @@ <h3 id="determine-requests-referrer">
...
+                  URL</a>, then return <code>no referrer</code>.
+                </li>
+              </ol>
+            </li>
+            <li>Return |referrerOrigin|.
+          </ol>
+        </dd>
+
+        <dt><a>"<code>safe-origin-when-cross-origin</code>"</a></dt>
+        <dt>the empty string</dt>
```
fmarier: This line should not be there because the empty string (i.e. the default policy state) maps to `no-referrer-when-downgrade`.

index.src.html
```diff
@@ -699,6 +757,53 @@ <h3 id="determine-requests-referrer">
...
+        <dd>
+          <ol>
+            <li>
```
fmarier: If I'm reading this correctly, the algorithm contained in this patch is:

```
if httpsPage and httpResource:
    if page and resource are same-origin:
        return fullURL
    else:
        return origin
else:
    return no-referrer
```

That's not quite right, I would suggest this instead:

```
1. If request is a same-origin request, then return referrerURL.
2. If environment is not null:
    1. If environment is TLS-protected and the origin of request’s current
       URL is not an a priori authenticated URL, then return no referrer.
3. Otherwise, return referrerOrigin.
```

domenic:
> We also want this upstreamed in Fetch and HTML.

A pull request would be welcome in both cases, I'm sure! Fetch I believe is pretty easy; just add the enum values to https://fetch.spec.whatwg.org/#referrer-policies.

For HTML, the places to update are:

- The "The following terms are defined in _Referrer Policy_" section
- The "Referrer policy attributes" section
- The index where the referrerpolicy content attribute is listed

You can find them all in the source by Ctrl+Fing for "origin-when-cross-origin".

index.src.html
mozfreddyb: Coming in a following commit. Thanks.

index.src.html
mozfreddyb: Fallout from over-eager copypasta. Thanks!

index.src.html
```diff
@@ -332,6 +332,32 @@ <h3 dfn export id="referrer-policy-origin" oldids="referrer-policy-state-origin"
     priori</em> authenticated URLs</a>.
   </div>
 
+  <h3 dfn export id="referrer-policy-safe-origin" oldids="referrer-policy-state-safe-origin">"<code>safe-origin</code>"</h3>
+
+  The <a>"<code>safe-origin</code>"</a> policy sends the
+  <a lt="ASCII serialization of an origin">ASCII serialization</a> of the
+  <a>origin</a> of the <a>request client</a> when making requests from
```
estark37: missing "a" at the end of this line

index.src.html
```diff
@@ -332,6 +332,32 @@ <h3 dfn export id="referrer-policy-origin" oldids="referrer-policy-state-origin"
...
+
+  Requests from <a>TLS-protected</a> <a>request clients</a> to non-<a><em>a
+  priori</em> authenticated URLs</a>, on the other hand, will contain no
+  referrer information. A <code><a>Referer</a></code> HTTP header will not be
+  sent.
+
+  <div class="example">
+    If a document at <code>https://example.com/page.html</code> sets a policy of
+    <a>"<code>safe-origin</code>"</a>, then navigations to
+    <code>https://<strong>not</strong>.eaxmple.com</code> would send a
```
estark37: "example" is misspelled as "eaxmple" here

index.src.html
```diff
@@ -332,6 +332,32 @@ <h3 dfn export id="referrer-policy-origin" oldids="referrer-policy-state-origin"
...
+    <a><code>Referer</code></a> header with a value of
+    <code>https://example.com/</code>. This explicitly excludes navigations to
+    URLs that are not <a><em>a priori</em> authenticated URLs</a>.
+    Navigations to <code><strong>http://not</strong>.example.com</code> would
+    send no <a><code>Referer</code></a> header.
```
estark37: Could you please include another example of what happens with requests from http://example.com?

index.src.html
```diff
@@ -332,6 +332,32 @@ <h3 dfn export id="referrer-policy-origin" oldids="referrer-policy-state-origin"
...
+  <a>TLS-protected</a> <a>environment settings object</a> to a
+  <a><em>a priori</em> authenticated URL</a>, and requests from
+  <a>request clients</a> which are <em>not</em> <a>TLS-protected</a> to any
+  <a>origin</a>.
+
+  Requests from <a>TLS-protected</a> <a>request clients</a> to non-<a><em>a
+  priori</em> authenticated URLs</a>, on the other hand, will contain no
+  referrer information. A <code><a>Referer</a></code> HTTP header will not be
+  sent.
+
```
estark37: Ok, is the following a correct summary of `safe-origin`?
- same-origin requests always include the full URL as the referrer
- https -> http requests have no referrer
- https -> https requests (cross-origin) have the origin as the referrer
- http -> http/https (cross-origin) have the origin as the referrer

Is that right? I find this section's text a little hard to understand as is, so want to make sure I understand the policy to see if I can figure out a way to make it easier to understand.

index.src.html
```diff
@@ -869,9 +980,14 @@ <h3 id="information-leakage">Information Leakage</h3>
   <a>"<code>unsafe-url</code>"</a> might leak the origin and the URL of
   a secure site respectively via insecure transport.
 
-  Those two policies are include in the spec nevertheless to lower the friction
+  Those two policies are included in the spec nevertheless to lower the friction
   of sites adopting secure transport.
 
+  Authors can use the policy states <a>"<code>safe-origin</code>"</a>,
+  <a>"<code>safe-origin-when-cross-origin</code>"</a> and
+  <a>"<code>no-referrer</code>"</a> to prevent sensitive information from
```
estark37: This section seems a little misleading, since I think every policy except `unsafe-url` could be used in some way to prevent sensitive information from leaking. For example, `origin` is a policy that can be used to protect capability URLs. Are there particular privacy properties of `safe-origin` and `safe-origin-when-cross-origin` that you're trying to call out here?

estark37:
Thanks @mozfreddyb and @fmarier! I left a few comments and questions inline. I'm not totally in love with the `safe-` naming scheme, but I can't think of any better ideas either. :( My concern is that `safe-` may make it sound universally safe when it isn't; in some scenarios, leaking your origin to other (https) origins might not actually be safe at all.

estark37:
Also, I'm still a little confused about the use case for these policies. HTTPS doesn't really stop a network eavesdropper from learning the origin, usually, so I don't see that a site gets all that much value from using `safe-origin` over `origin`.

index.src.html
fmarier: > Ok, is the following a correct summary of `safe-origin`?
> - same-origin requests always include the full URL as the referrer

No. Just like with `origin`, same-origin requests don't get the path.

> - https -> http requests have no referrer

Correct. That's the only difference when compared to `origin`.

> - https -> https requests (cross-origin) have the origin as the referrer
> - http -> http/https (cross-origin) have the origin as the referrer

Correct.

fmarier:
> I'm not totally in love with the `safe-` naming scheme, but I can't think of any better ideas either. :(

Following on [another recent rename](https://lists.w3.org/Archives/Public/public-webappsec/2016Jun/0047.html), how about `strict-origin` and `strict-origin-when-cross-origin`?

It feels more appropriate since "strict" sounds more secure without implicitly calling the other policy states "unsafe".

If you like that better, I'm happy to rename it in the pull request.

fmarier:
> Also, I'm still a little confused about the use case for these policies. HTTPS doesn't really stop a network eavesdropper from learning the origin, usually, so I don't see that a site gets all that much value from using safe-origin over origin.

In the case of a network attacker that sees all of the user's traffic (e.g. their ISP), you're right, this doesn't achieve much.

I can think of two examples where it makes a difference:

1. A user visits `https://example.com`.
2. `example.com` uses the `origin` policy.
3. An attacker can see all incoming traffic to `wikileaks.org`.
4. User follows a link from `https://example.com` to `http://wikileaks.org`.
5. Attacker finds out a small piece of the user's browsing history (example.com) through the Referrer.

The second one:

1. A user visits `https://example.com`.
2. `example.com` uses the `origin` policy.
3. `example.com` loads images from `http://catgifs.com` and ads from `https://ads.com`.
4. The image hosting service and the ad server both learn about the user's visit to `example.com`.

In these examples, both the attacker and the image hosting service learn about `example.com`. They wouldn't learn that under the default policy.

We recently had [a request](https://lists.w3.org/Archives/Public/public-webappsec/2016Jun/0060.html) for a policy like `safe-origin-when-cross-origin` and I would also like to be able to tell developers wanting to use `origin` that they should instead use `safe-origin` unless they specifically need to send the origin to HTTP sites.

index.src.html
fmarier: Fixed: https://github.com/w3c/webappsec-referrer-policy/pull/56/commits/9a2111ae3404db6600d32a57d9389b36c1333729

index.src.html
fmarier: Fixed: https://github.com/w3c/webappsec-referrer-policy/pull/56/commits/9a2111ae3404db6600d32a57d9389b36c1333729

index.src.html
fmarier: Added in https://github.com/w3c/webappsec-referrer-policy/pull/56/commits/259bccfd84e1d474d9f50447ef25d229221b8735

index.src.html
fmarier: I took a stab at rephrasing this in https://github.com/w3c/webappsec-referrer-policy/pull/56/commits/2bce42c18e149734ab2bfb68993393cfcbc8f74f

fmarier:
Given that @mozfreddyb is out and that I couldn't push more commits to his personal repo, I created a new pull request for this: #56.

Let's move the discussion there.

PR #56, opened 2016-06-29
Add strict-origin and strict-origin-when-cross-origin policy states (fixes #50)

### LINKED ISSUE #50: Add the non-downgrading versions of origin and origin-when-cross-origin

fmarier:
Since @mozfreddyb  is out of the office until September, I'll take over his patch and so this pull request supersedes #52.

The three commits I added should address the issues that @estark37 found.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -332,6 +332,32 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
     priori</em> authenticated URLs</a>.
   </div>
 
+  <h3 dfn export id="referrer-policy-safe-origin" oldids="referrer-policy-state-safe-origin">"<code>safe-origin</code>"</h3>
+
+  The <a>"<code>safe-origin</code>"</a> policy sends the
+  <a lt="ASCII serialization of an origin">ASCII serialization</a> of the
+  <a>origin</a> of the <a>request client</a> when making requests from
+  <a>TLS-protected</a> <a>environment settings object</a> to a
+  <a><em>a priori</em> authenticated URL</a>, and requests from
+  <a>request clients</a> which are <em>not</em> <a>TLS-protected</a> to any
+  <a>origin</a>.
+
+  Requests from <a>TLS-protected</a> <a>request clients</a> to non-<a><em>a
+  priori</em> authenticated URLs</a>, on the other hand, will contain no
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -332,6 +332,42 @@ <h3 dfn export id="referrer-policy-origin" oldids="referrer-policy-state-origin"
     priori</em> authenticated URLs</a>.
   </div>
 
+  <h3 dfn export id="referrer-policy-safe-origin" oldids="referrer-policy-state-safe-origin">"<code>safe-origin</code>"</h3>
+
+  The <a>"<code>safe-origin</code>"</a> policy sends the
+  <a lt="ASCII serialization of an origin">ASCII serialization</a> of the
+  <a>origin</a> of the <a>request client</a> when making requests from a
+  <a>TLS-protected</a> <a>environment settings object</a> to a
+  <a><em>a priori</em> authenticated URL</a>, and requests from
```
estark37: Would you be okay with changing "..., and requests from" to "..., and when making requests from"? To me that makes it a little easier to parse.

estark37:
@fmarier thanks for updating this and for the examples in https://github.com/w3c/webappsec-referrer-policy/pull/52#issuecomment-229515024.

I'd be fine with `strict-` and I like that better than `safe-`. Mostly for completeness, I'll suggest that we could be more explicit and do something like `origin-when-not-downgrading` and `origin-when-cross-origin-not-downgraded` or something like that... but that's super long and also grammatically awkward so `strict-` is probably better.

fmarier:
> Would you be okay with changing "..., and requests from" to "..., and when making requests from"? To me that makes it a little easier to parse.

I've turned that bit into a bullet list. Do you think that's easier to read?

The actual wording is still pretty dense. It would be great if we could say "HTTPS page loading HTTP resource" in less words than what we currently have.

estark37:
Thanks @fmarier, merged! I think the bulleted list makes it a lot easier to read.

estark37:
And thanks @mozfreddyb too.

Issue #43, opened 2016-04-29
HTML integration plan


domenic:
I got started on HTML integration in https://github.com/whatwg/html/pull/1153 and https://github.com/w3c/webappsec-referrer-policy/pull/42, but there is still much to do. Here are the issues I see:

- [x] HTML's definition of "environment settings object" should explicitly get a "referrer policy" field, instead of having it be patched in in "Core concepts" (https://github.com/whatwg/html/pull/1205)
- [x] The monkeypatch in "Integration with HTML", regarding creating Documents or WorkerGlobalScopes, should be integrated into HTML (https://github.com/whatwg/html/pull/1205)
- [x] The monkeypatches in "Implicit delivery" should be integrated into HTML (https://github.com/whatwg/html/pull/1205)
- [x] HTML has removed "API referrer source" in the course of Fetch integration. @annevk, how should this be changed?
- [x] All places where the `referrerpolicy` content attribute applies should explicitly pass along the referrer policy when making the appropriate fetch (#40), similar to what is done with the `crossorigin` content attribute today.

annevk:
"API referrer source" is now basically the environment settings object itself rather than a field on it, passed to fetch as request's client. (Or you can set an explicit referrer through request's referrer.)

domenic:
@annevk with that in mind do you have a suggestion for how to fix https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer step 3.3.2? It doesn't seem obvious to me.

annevk:
I think for a worker you'd just inspect the worker itself for the referrer policy. That would be analogous to CSP. What it states now seems wrong, but I might be missing something?

domenic:
I was hoping for some explicit text, or a pull request on top of my #42 :)

estark37:
@domenic did you happen to already have a plan for fixing the "API referrer source" reference in https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer? I wonder if it should be the creation URL of the environment settings object...

domenic:
I don't have a plan; I was hoping @annevk would give more concrete suggestions. But maybe this isn't so bad after all...

> I wonder if it should be the creation URL of the environment settings object...

I believe that is correct, for the worker case. For the document case, the current text seems correct, although I'm not sure on the difference between the responsible browsing context's active document, and the responsible document.

The algorithm looks like it could be cleaned up a decent bit. Here is my draft to replace the current step 3:

[...]

Issue #31, opened 2016-04-20
Srcdoc referrer policies don't interact well with <meta referrer>


bzbarsky:
The way the spec is written right now, a srcdoc iframe snapshots its owner document's referrer policy at the point when it creates its browsing context.  If that parent document's referrer policy then changes due to a `<meta referrer>`, the srcdoc's referrer policy will NOT change.  Is that the desired behavior?

estark37:
I think this is desired behavior and was resolved in https://github.com/whatwg/html/pull/1205, is that correct? (@domenic)

domenic:
Right, I think we concluded that, with some of the extra formalization performed in https://github.com/whatwg/html/pull/1205, the behavior described here is the desired one.

bzbarsky:
Why is that the desired behavior?  The referrer is dynamically grabbed from the container document, but the referrer policy is snapshotted?  Seems to me they should both be snapshots or both be dynamic.

domenic:
Hmm, where is the referrer grabbed dynamically from the container document?

> The document's referrer is a string (representing a URL) that can be set when the Document is created.

- https://html.spec.whatwg.org/#initialise-the-document-object set them together (steps 7 and 9) 
- https://html.spec.whatwg.org/#creating-a-new-browsing-context set them together (steps 9 and 10)

bzbarsky:
https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer step 3 substep 3 subsubsteps 2 and 3.

bzbarsky:
And note that https://html.spec.whatwg.org/#initialise-the-document-object does not set the referrer for srcdoc documents, because in step 9 the "If resource is a request" precondition is false: a srcdoc has a response but not a request.  See https://html.spec.whatwg.org/#process-the-iframe-attributes

So for srcdoc documents the "document's referrer" is never set right now per spec, as far as I can tell.  Luckily, https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer never tries to use it for a srcdoc document.

domenic:
I see, you were talking about the referrer/referrer policy at fetching time, not document creation time. So about the referrer of requests initiated inside that document, not about `document.referrer`.

After thinking about this a bit more, I appreciate the asymmetry you are referring to. Roughly, when performing a request from "inside" an iframe srcdoc document:

- We determine the referrer by saying "oh, the URL is meaningless (it's about:srcdoc); let's look up to the parent".
[...]

PR #58, opened 2016-07-14
Fix referrer calculation to not refer to "API referrer source"

domenic:
This was discussed in #43. It doesn't necessarily help with #31, but is a good stage-setting for whatever we do there eventually...

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -49,6 +49,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: forced sandboxing flag set
       text: auxiliary browsing context
       text: browsing context
+      text: browsing context; for: Document; url: concept-document-bc
       text: parent browsing context
       text: ancestor browsing context
       text: browsing context container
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -683,74 +684,53 @@ <h3 id="determine-requests-referrer">
   associated with it, as detailed in the following steps, which return
   either <code>no referrer</code> or a URL:
 
-  Note: If Fetch is performing a navigation in response to a link of type
-  <code><a>noreferrer</a></code>, then <var>request</var>'s
```
estark37: Why delete this? I think it's still true/useful.

estark37: Oh, never mind, I just saw you put it below instead.

estark37:
Thanks @domenic. Just trying to figure out what's going on with the Travis build...

index.src.html
domenic: Yeah, I thought it would be nice to put it in the switch, because when I was mentally reading that part of the algorithm, I was like, "what about if it's "`no-referrer`"?" and it took me a while to look up and find this note.

PR #60, opened 2016-07-15
Update Integration with Fetch section

estark37:
The description of when a referrer policy is set on redirect referred to
a non-existent step in the HTTP-redirect fetch algorithm. The
corresponding Fetch change is in
https://github.com/whatwg/fetch/pull/335.

The section about when "Determine request's referrer" is called was also
inaccurate.


estark37:
@jdm do you think you could take a look and tell me if this looks right?

jdm:
Looks good!

Issue #29, opened 2016-04-20
Use of "active document" in determining the referrer looks wrong


bzbarsky:
https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer step 3 substep 3 subsubstep 1 says:

>  Let document be the Document object of the active document of the browsing context of environment’s responsible browsing context. 

What this means is that if a load happens in a document that's _not_ active, the referrer will be set to whatever document is active in the same browsing context, which has nothing whatsoever to do with where the load is happening.  I'm not sure what this bit is trying to do by using "active document", but this doesn't seem right.

domenic:
After #58 this has been slightly rejiggered so that the relevant text is now

> Let _document_ be the `Document` object of the active document of the browsing context of _environment_’s responsible browsing context. 

It sounds like the correct fix here is to use _environment_'s global object's associated Document? (_environment_'s global object is guaranteed to be a Window in this step).

PR #63, opened 2016-08-17
Fix use of "active document" in determining referrer

estark37:
"Determine request's referrer" now uses the environment's global
object's associated Document instead of the active document of its
browsing context. Fixes https://github.com/w3c/webappsec-referrer-policy/issues/29.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -45,6 +45,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: browsers.html
       text: opaque origin; url: concept-origin-opaque
       text: active document
+      text: associated Document; url: concept-document-window
       text: parse a sandboxing directive
       text: forced sandboxing flag set
       text: auxiliary browsing context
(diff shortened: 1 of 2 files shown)


estark37:
@domenic does this look correct to you?

index.src.html
```diff
@@ -703,8 +704,8 @@ <h3 id="determine-requests-referrer">
               object</a> is a {{Window}} object, then
 
               <ol>
-                <li>Let |document| be the <a>active document</a> of
-                |environment|'s <a>responsible browsing context</a>.</li>
+                <li>Let |document| be the <a>associated Document</a> of
```
domenic: Nit: should be `<code>Document</code>`

index.src.html
```diff
@@ -703,8 +704,8 @@ <h3 id="determine-requests-referrer">
...
+                |environment|'s <a>global object</a>.</li>
```
domenic: Should be `<a for="environment settings object">global object</a>`

domenic:
Looks good with some editorial tweaks. Yay!

estark37:
Thanks @domenic!

PR #64, opened 2016-08-18
Copy referrer policy concept and enum from Fetch

### LINKED ISSUE #62: Take over "referrer policy" concept and ReferrerPolicy enum from Fetch
The setup I created to support referrer policies in the `fetch()` API makes introducing new values rather complicated as three specifications have to be updated.

If Referrer Policy includes https://fetch.spec.whatwg.org/#referrer-policies Fetch could just refer to Referrer Policy for that. Maybe HTML could do a similar thing, with HTML defining the attribute and Referrer Policy the value space.

estark37:
By keeping the definition in REFERRER, adding new policy values requires
only changing one spec, instead of three.

Fixes #62

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -21,7 +21,6 @@ spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
     text: fetching
     text: request; url: concept-request
     text: response; url: concept-response
-    text: referrer policy; url: concept-referrer-policy
     text: request client; url: concept-request-client
   type: interface
     text: Request
@@ -93,7 +92,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
     urlPrefix: workers.html
       text: running a worker; url: run a worker
     urlPrefix: dom.html
-      text: referrer policy; url: concept-document-referrer-policy; for: Document
+      text: document referrer policy; url: concept-document-referrer-policy; for: Document
   type: interface
     urlPrefix: dom.html
       text: Document
@@ -132,7 +131,6 @@ spec: RFC2616; urlPrefix: https://tools.ietf.org/html/rfc7231
 
 <pre class="link-defaults">
 spec: HTML; type: element; text: link;
-spec: FETCH; type: dfn; text: referrer policy; for: /;
 </pre>
 
 <!--
@@ -233,8 +231,29 @@ spec: FETCH; type: dfn; text: referrer policy; for: /;
 <section>
   <h2 id="referrer-policies" oldids="referrer-policy-states">Referrer Policies</h2>
 
-  Each possible <a>referrer policy</a>, besides the empty string, is explained
-  below. A detailed algorithm for evaluating their effect is given in the
+  A <dfn title="concept-referrer-policy" data-export>referrer policy</dfn>
+  is the empty string, "<code>no-referrer</code>",
+  "<code>no-referrer-when-downgrade</code>", "<code>same-origin</code>",
+  "<code>origin</code>", "<code>strict-origin</code>",
+  "<code>origin-when-cross-origin</code>",
+  "<code>strict-origin-when-cross-origin</code>", or
+  "<code>unsafe-url</code>".
+
+<pre class=idl>enum <dfn data-dfn-type=enum data-export>ReferrerPolicy</dfn> {
+
+  "",
+  "no-referrer",
+  "no-referrer-when-downgrade",
+  "same-origin",
+  "origin",
+  "strict-origin",
+  "origin-when-cross-origin",
+  "strict-origin-when-cross-origin",
+  "unsafe-url"
+};</pre>
+
+  Each possible <a>referrer policy</a> is explained below. A detailed
+  algorithm for evaluating their effect is given in the
   <a section href="#integration-with-fetch"></a> and
   <a section href="#algorithms"></a> sections.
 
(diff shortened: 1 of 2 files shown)


estark37:
@annevk @domenic could you please take a look? Thanks!

index.src.html
```diff
@@ -233,8 +231,29 @@ <h2 id="terms">Key Concepts and Terminology</h2>
...
+  A <dfn title="concept-referrer-policy" data-export>referrer policy</dfn>
+  is the empty string, "<code>no-referrer</code>",
+  "<code>no-referrer-when-downgrade</code>", "<code>same-origin</code>",
+  "<code>origin</code>", "<code>strict-origin</code>",
+  "<code>origin-when-cross-origin</code>",
+  "<code>strict-origin-when-cross-origin</code>", or
+  "<code>unsafe-url</code>".
+
+<pre class=idl>enum <dfn data-dfn-type=enum data-export>ReferrerPolicy</dfn> {
+
```
domenic: Extra blank line seems a bit ugly...

domenic:
Looks good with a nit; @annevk should probably sign off, but seems pretty un-problematic to me :). I'll go work on the HTML side now.

index.src.html
estark37: I don't understand why and feel pretty dumb, but for some reason a line gets removed by bikeshed, so this comes out looking like you'd want. (See line 1252 of index.html.)

domenic: Oh wow, I have no idea why that would happen either. Oh well! Sorry for not checking the output.

mikewest: I'd suggest going with formatting more like:

```
<pre class="idl">
  enum ReferrerPolicy {
    ...
  };
</pre>
```

Bikeshed will automatically remove the leading whitespace, and this is simpler to read.

Also, you should be able to drop the `<dfn>`, as Bikeshed ought to automatically process the IDL to produce the expected markup.

index.src.html
```diff
@@ -475,10 +494,11 @@ <h3 dfn export id="referrer-policy-empty-string">The empty string</h3>
...
     Given a HTML <{a}> element without any declared <{a/referrerpolicy}>
-    attribute, its referrer policy is "". Thus, navigation requests initiated
-    by clicking on that <{a}> element will be sent with the <a
-    for="Document">referrer policy</a> of the <{a}> element's <a>node
-    document</a>. If that {{Document}} has "" as its referrer policy, the
+    attribute, its referrer policy is "". Thus, navigation requests
+    initiated by clicking on that <{a}> element will be sent with
+    the <a for="Document" lt="document referrer policy">referrer
+    policy</a> of the <{a}> element's <a>node document</a>. If that
+    {{Document}} has "" as its referrer policy, the
```
annevk: I think it would be clearer to write out "" as "the empty string" but that could be follow up since that was already an issue.

annevk:
This looks good. I did notice that some of the references to Fetch are wrong. E.g., instead of pointing to https://fetch.spec.whatwg.org/#concept-request-referrer-policy the document points to https://fetch.spec.whatwg.org/#concept-referrer-policy for "request's referrer policy" and that now becomes an internal reference while it's actually attempting to reference a concept that Fetch will continue to define.

index.src.html
```diff
@@ -475,10 +494,11 @@ <h3 dfn export id="referrer-policy-empty-string">The empty string</h3>
 
   <div class="example">
     Given a HTML <{a}> element without any declared <{a/referrerpolicy}>
-    attribute, its referrer policy is "". Thus, navigation requests initiated
-    by clicking on that <{a}> element will be sent with the <a
-    for="Document">referrer policy</a> of the <{a}> element's <a>node
-    document</a>. If that {{Document}} has "" as its referrer policy, the
+    attribute, its referrer policy is "". Thus, navigation requests
+    initiated by clicking on that <{a}> element will be sent with
+    the <a for="Document" lt="document referrer policy">referrer
```
mikewest: Why do you need to link this to `document referrer policy`? The `for="Document"` should be enough to disambiguate (and if it isn't, we should tell Tab. :) ).

index.src.html
estark37: Aha, worked like a charm, thanks @mikewest

index.src.html
estark37: Done.

index.src.html
estark37: I think this was because Document referrer policy was listed as type `dfn` instead of `interface`; I fixed that above and now `{{Document/referrer policy}}` works.

estark37:
Thanks for the comments @annevk, @domenic, @mkwst. I made some tweaks. And I updated some `concept-referrer-policy` references to `concept-request-referrer-policy` as @annevk suggested.

index.src.html
```diff
@@ -91,11 +92,10 @@
       text: creation URL
     urlPrefix: workers.html
       text: running a worker; url: run a worker
-    urlPrefix: dom.html
-      text: document referrer policy; url: concept-document-referrer-policy; for: Document
   type: interface
     urlPrefix: dom.html
       text: Document
+      text: referrer policy; url: concept-document-referrer-policy; for: Document
```
domenic: No, that's not right; interfaces are Web IDL things declared as `interface X`. I'm surprised Bikeshed even lets you use an "interface" with a space in the name...

estark37: Darn, I must be totally confused. So should this be a `type: dfn`, with the link as `<a lt="document referrer policy">referrer policy</a>` and no `for` attribute?

domenic: I think it should be under the dfns, but with

```
    urlPrefix: dom.html
      text: referrer policy; url: concept-document-referrer-policy; for: Document
```

and then you can do `<a for="Document">referrer policy</a>`.

If that doesn't work we likely have a Bikeshed bug on our hands :(

estark37: When I do that, it gets confused about multiple definitions for 'referrer policy' and chooses to link all the `<a>referrer policy</a>`'s in the spec to the HTML Document definition. Is there any way to tell it that I want the others to link to this spec's referrer policy dfn? I think I tried to put something in the link-defaults section but it didn't work; I'll play with that again...

domenic: This is sounding familiar... https://github.com/tabatkins/bikeshed/issues/786 I guess. Ugh.

Maybe as a workaround

```
    urlPrefix: dom.html
      text: document referrer policy; url: concept-document-referrer-policy; for: Document
```

with `<a for="Document" lt="document referrer policy">referrer policy</a>` will work, although if you do that I'd add a comment about it being a hack.

estark37: Gotcha, that'll work for now. Added a comment.

estark37:
@annevk and @domenic, do you have any more comments before I merge this?

domenic:
Looks good to me :)

estark37:
I'm going to go ahead and merge, but @annevk please do let me know if you have any more comments and I'll address as a follow-up.

Issue #53, opened 2016-06-11
What is the motivation for "legacy keywords" in the header?


domenic:
Why does the Referrer-Policy header allow "never", "default", and "always"?

I was under the impression it was because there were shipping implementations that allowed them. However, https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/Umj9iVRJM70 indicates that nobody has implemented the header yet, with Blink being the first.

At this point my best guess is that a few user agents have already shipped `<meta name="referrer">`, which allows the legacy keywords, and that the desire is that the Referrer-Policy header be consistent with the `<meta>` element?

I don't think that's a very strong argument. Mainly because we're using `<meta name="referrer">`, not `<meta http-equiv="Referrer-Policy">`, so there is no reason to say that the meta element should be consistent with the HTTP header.

As such, it seems better to me to say that the HTTP header matches the `referrerpolicy=""` content attribute in disallowing the legacy values, instead of saying it must match `<meta>` in allowing them.

estark37:
I buy your argument that it shouldn't support the legacy keywords, but I don't see that it actually does. Where are you looking? I'm looking at https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-header which doesn't list "never" or "default" or "always" in the policy-token values.

estark37:
By the way, @fmarier, can you confirm that Firefox hasn't implemented the Referrer-Policy header yet? I wasn't 100% sure if you had or not.

fmarier:
> can you confirm that Firefox hasn't implemented the Referrer-Policy header yet?

Correct, we haven't implemented it yet: https://bugzilla.mozilla.org/show_bug.cgi?id=1264164

estark37:
Doh, don't know how I failed to find that in my bug search, thanks @fmarier!

domenic:
Well, https://w3c.github.io/webappsec-referrer-policy/#parse-referrer-policy-from-header links to https://w3c.github.io/webappsec-referrer-policy/#determine-policy-for-token which processes the legacy keywords. I'm not sure what the grammar production is used for...

estark37:
Ahh, ok, I see, you're right!

estark37:
I think that determine-policy-for-token is only used for Referrer-Policy headers, so perhaps we can just delete the legacy keywords from that algorithm. (And maybe just delete the grammar too.)

annevk:
[...]

Issue #65, opened 2016-08-19
`Referrer-Policy` header syntax is incompatible with JSON.


mikewest:
In the conversation on https://github.com/w3c/webappsec-referrer-policy/pull/14, it seemed like there was agreement (or at least, no objection) to the notion of retaining forward-compatibility with a more robust structure based on JSON. The current syntax doesn't. Was that an intentional decision?

That is, `Referrer-Policy: bareword` would make a JSON parser sad, whereas `Referrer-Policy: "quoted-keyword"` would not. I think the quoted version makes more sense, given the direction we seem to be going with header parsing. WDYT, @estark37?

/cc @fmarier @annevk

estark37:
Wow, that whole part of the conversation seems to have completely dropped off my radar. Don't know what happened there but thanks for reminding me about it! I agree we should do this.

I'll put a PR together. (And we'll have to update the WPT and Chrome's implementation too.)

annevk:
Firefox implements this header according to @franziskuskiefer and it'll hit stable in Firefox 50 (two releases from now it seems). Not sure if we still have room to change this around.

mikewest:
If no one's using the current syntax (and I don't see how they could be using it, since no one is shipping it yet), then it seems like there's wiggle room. It might be more or less _comfortable_ depending on how close vendors are to shipping, but until there's real usage, it seems worth the mild discomfort to make the new syntax forward-compatible with a more structured encoding in the future.

dveditz:
Is there any other HTTP header that has quoted values? Developers are going to leave off the quotes out of habit if nothing else leading to lots of broken policies or else implementations being "helpful" and interpreting them both ways.

The spec will have to be clear what happens then. An error and no restriction (default behavior) on referrer? An error and then fallback to the safe-and-hopefully-noticed no referrer at all?

estark37:
@dveditz I would say unquoted values should be ignored like any other invalid policy, maybe with a console warning. (Can't remember if Chrome already has a console message for invalid policies or not.)

Also this is kind of ugly, but if we don't change the values to be quoted now, it's conceivable that we could still introduce a JSON-based syntax later with something like `Referrer-Policy: custom; {...}`, where there's an unquoted keyword to indicate that a JSON policy follows. Or something.

jeisinger:
we discussed this at the TPAC meeting today and concluded to stick with the bare words for now.

jeisinger:
closing as per previous comment

Issue #46, opened 2016-05-04
Clarify that "Unknown Policy Values" applies to Document policies


estark37:
http://w3c.github.io/webappsec-referrer-policy/#unknown-policy-values describes why unknown policy values are ignored and gives an example of how this can be used to deploy new policy values. This should be clarified that it only applies to Document policies delivered via HTTP header or meta tag, not to `referrerpolicy` attributes (which can only have one token). Authors can use feature detection to deploy new policy values in `referrerpolicy` attributes.

(see discussion on https://github.com/whatwg/html/pull/1153)

PR #66, opened 2016-08-25
Use referrer policy definition in parsing Referrer-Policy headers

estark37:
As suggested in https://github.com/w3c/webappsec-referrer-policy/issues/53, instead of
using "Determine token's policy" to parse Referrer-Policy headers, we
can simply use the referrer policy definition. This also makes it more
explicit that legacy keywords are not supported in Referrer-Policy
headers.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -674,10 +674,13 @@ spec: HTML; type: element; text: link;
       Let |policy| be the empty string.
     </li>
     <li>
-      For each <var>token</var> in <var>policy-tokens</var>,
-      if <var>token</var> is a <a>referrer policy</a>
-      and <var>token</var> is not the empty string, then
-      set <var>policy</var> to <var>token</var>.
+      For each |token| in |policy-tokens|, if |token| is a <a>referrer
+      policy</a> and |token| is not the empty string, then set |policy|
+      to |token|.
+
+	  Note: This algorithm loops over multiple policy values to allow
+	  deployment of new policy values with fallbacks for older user
+	  agents, as described in [[#unknown-policy-values]].
     </li>
     <li>
       Return |policy|.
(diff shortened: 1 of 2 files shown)


estark37:
@domenic would you be able to review this for me? Doesn't yet use the JSON quoted string as in https://github.com/w3c/webappsec-referrer-policy/issues/65.

index.src.html
```diff
@@ -665,15 +667,17 @@ <h3 id="parse-referrer-policy-from-header">
...
+      |response|'s <a>header list</a>.
     </li>
     <li>
       Let |policy| be the empty string.
     </li>
     <li>
-      For each <var>token</var> in <var>policy-tokens</var>, execute
-      [[#determine-policy-for-token]] on <var>token</var> and
-      set <var>policy</var> to the result if it is not the empty string.
+      For each <var>token</var> in <var>policy-tokens</var>,
```
domenic: Maybe while you're here you might want to consistify the algorithm between `|variable|` and `<var>variable</var>` :)

index.src.html
```diff
@@ -665,15 +667,17 @@ <h3 id="parse-referrer-policy-from-header">
...
+      if <var>token</var> is a <a>referrer policy</a>
+      and <var>token</var> is not the empty string, then
+      set <var>policy</var> to <var>token</var>.
     </li>
```
domenic: Maybe as part of fixing #46, this step should include a pointer to https://w3c.github.io/webappsec-referrer-policy/#unknown-policy-values, since that's why the algorithm is doing a loop.

domenic:
LGTM, just a few potential notes.

domenic:
Oh, did you also want to delete "Determine token's Policy" in this commit?

estark37:
Thanks. For deleting "Determine token's Policy", I was thinking we should first make the change in HTML to translate the legacy keywords there?

domenic:
Yeah, that sounds like a proper thing to do :)

index.src.html
estark37: Done

index.src.html
estark37: Done

domenic:
LGTM!

PR #67, opened 2016-08-25
Remove "Determine policy for token"

### LINKED ISSUE #53: What is the motivation for "legacy keywords" in the header?

estark37:
The HTML meta element now translates legacy keywords into real referrer
policies, so this algorithm is no longer needed.

While removing a reference to it in "Unknown Policy Values", I also
clarified that this section applies to Document policies, not
'referrerpolicy' attributes.

Fixes #53, #46.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1153,7 +1153,6 @@
       <li><a href="#set-requests-referrer-policy-on-redirect"><span class="secno">7.2</span> <span class="content"> Set <var>request</var>’s referrer policy on redirect </span></a>
       <li><a href="#determine-requests-referrer"><span class="secno">7.3</span> <span class="content"> Determine <var>request</var>’s Referrer </span></a>
       <li><a href="#strip-url"><span class="secno">7.4</span> <span class="content"> Strip <var>url</var> for use as a referrer </span></a>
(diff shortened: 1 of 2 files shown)


estark37:
note to self: do not merge until @domenic merges https://github.com/whatwg/html/pull/1714

estark37:
@domenic now that your HTML PR is merged, does this look okay to you?

domenic:
Looks great!!

PR #68, opened 2016-09-22
Add a section about CSS and referrers (#5)

jeisinger:
/cc @estark37 @annevk @domenic

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -655,6 +655,27 @@ spec: HTML; type: element; text: link;
   with W3C HTML5, those would need to be copied from [[HTML]].
 </section>
 
+<section class="informative">
+  <h2 id="integration-with-css">Integration with CSS</h2>
+
+  <em>This section is not normative.</em>
+
+  The CSS Standard does not specify how it fetches resources referenced from
+  stylesheets. However, implementations use the URL a given stylesheet was
+  loaded from as referrer for the requests for resources form that stylesheet.
+
+  Implementations should keep track of a referrer policy for each stylesheet
+  that should be used to create requests for resources from the respective
+  stylesheet. The referrer policy should be inherited from the entity that
+  triggered the load of the stylesheet, i.e., a document, or another
+  stylesheet.  If the network response of the stylesheet included an
+  `<code>Referrer-Policy</code>` header, the header's policy should take
+  precedence.
+
+  Both, the value of the referrer and the value of the referrer policy should be
+  captured when the request for the stylesheet is created.
+</section>
+
 <section>
   <h2 id="algorithms">Algorithms</h2>
 
--- a/index.src.html
+++ b/index.src.html
@@ -655,7 +655,7 @@ spec: HTML; type: element; text: link;
   with W3C HTML5, those would need to be copied from [[HTML]].
 </section>
 
-<section class="informative">
+<section>
   <h2 id="integration-with-css">Integration with CSS</h2>
 
   The CSS Standard does not specify how it fetches resources referenced from
@@ -675,10 +675,8 @@ spec: HTML; type: element; text: link;
   requests for resources referenced from the stylesheet.
 
   For inline stylesheets, the referrer policy is the containing {{Document}}'s
-  referrer policy.
-
-  Both the value of the referrer and the value of the referrer policy should be
-  captured at the time a given request is created.
+  referrer policy.  Both the value of the referrer and the value of the
+  referrer policy should be captured at the time a given request is created.
 </section>
 
 <section>
--- a/index.src.html
+++ b/index.src.html
@@ -674,7 +674,8 @@ spec: HTML; type: element; text: link;
   declared <{link/referrerpolicy}>, this referrer policy will not affect the
   requests for resources referenced from the stylesheet.
 
-  For inline stylesheets, the referrer policy is the containing {{Document}}'s
+  For inline stylesheets, and styles applied via an <code>style</code>
+  attribute on an element, the referrer policy is the containing {{Document}}'s
   referrer policy.  Both the value of the referrer and the value of the
   referrer policy should be captured at the time a given request is created.
 </section>
--- a/index.src.html
+++ b/index.src.html
@@ -670,6 +670,10 @@ spec: HTML; type: element; text: link;
   <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
   `<code>Referrer-Policy</code>` header.
 
+  Note: If the stylesheet was loaded via a HTML <{link}> element with a
+  declared <{link/referrerpolicy}>, this referrer policy will not affect the
+  requests for resources referenced from the stylesheet.
+
   For inline stylesheets, the referrer policy is the containing {{Document}}'s
   referrer policy.
 
--- a/index.src.html
+++ b/index.src.html
@@ -658,22 +658,23 @@ spec: HTML; type: element; text: link;
 <section class="informative">
   <h2 id="integration-with-css">Integration with CSS</h2>
 
-  <em>This section is not normative.</em>
-
   The CSS Standard does not specify how it fetches resources referenced from
   stylesheets. However, implementations use the URL a given stylesheet was
-  loaded from as referrer for the requests for resources form that stylesheet.
+  loaded from as referrer for the requests for resources from that stylesheet.
 
   Implementations should keep track of a referrer policy for each stylesheet
   that should be used to create requests for resources from the respective
-  stylesheet. The referrer policy should be inherited from the entity that
-  triggered the load of the stylesheet, i.e., a document, or another
-  stylesheet.  If the network response of the stylesheet included an
-  `<code>Referrer-Policy</code>` header, the header's policy should take
-  precedence.
+  stylesheet. 
 
-  Both, the value of the referrer and the value of the referrer policy should be
-  captured when the request for the stylesheet is created.
+  For external stylesheets, the referrer policy should be
+  <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
+  `<code>Referrer-Policy</code>` header.
+
+  For inline stylesheets, the referrer policy is the containing {{Document}}'s
+  referrer policy.
+
+  Both the value of the referrer and the value of the referrer policy should be
+  captured at the time a given request is created.
 </section>
 
 <section>
--- a/index.html
+++ b/index.html
@@ -1421,7 +1421,7 @@ pre.idl.highlight { color: #708090; }
   <div class="head">
    <p data-fill-with="logo"><a class="logo" href="https://www.w3.org/"> <img alt="W3C" height="48" src="https://www.w3.org/StyleSheets/TR/2016/logos/W3C" width="72"> </a> </p>
    <h1>Referrer Policy</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2016-09-28">28 September 2016</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2016-09-29">29 September 2016</time></span></h2>
    <div data-fill-with="spec-metadata">
     <dl>
      <dt>This version:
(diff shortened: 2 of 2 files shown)


index.src.html
```diff
@@ -655,6 +655,27 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  <em>This section is not normative.</em>
+
+  The CSS Standard does not specify how it fetches resources referenced from
+  stylesheets. However, implementations use the URL a given stylesheet was
+  loaded from as referrer for the requests for resources form that stylesheet.
+
+  Implementations should keep track of a referrer policy for each stylesheet
+  that should be used to create requests for resources from the respective
+  stylesheet. The referrer policy should be inherited from the entity that
+  triggered the load of the stylesheet, i.e., a document, or another
```
jeisinger: @annevk pointed out that gecko uses the stylesheet's URL as referrer. I'd argue that the main point should be that the policy comes from the same place as the referrer. So if the stylesheet is the referrer, then the only way to get the referrer policy was via the referrer policy header.

index.src.html
```diff
@@ -655,6 +655,27 @@ <h2 id="integration-with-html">Integration with HTML</h2>
   with W3C HTML5, those would need to be copied from [[HTML]].
 </section>
 
+<section class="informative">
+  <h2 id="integration-with-css">Integration with CSS</h2>
+
+  <em>This section is not normative.</em>
```
domenic: I think the intention is more like, "This section is normative but we want it to go away when CSS gets its act together and specifies all this more precisely."

index.src.html
```diff
@@ -655,6 +655,27 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+
+  The CSS Standard does not specify how it fetches resources referenced from
+  stylesheets. However, implementations use the URL a given stylesheet was
+  loaded from as referrer for the requests for resources form that stylesheet.
```
domenic: s/form/from

index.src.html
estark37: In that argument, if there was no Referrer-Policy header served with the stylesheet, is the referrer policy the default? Or does it fall back to the document's policy?

index.src.html
```diff
@@ -655,6 +655,27 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+
+  Implementations should keep track of a referrer policy for each stylesheet
+  that should be used to create requests for resources from the respective
+  stylesheet. The referrer policy should be inherited from the entity that
+  triggered the load of the stylesheet, i.e., a document, or another
+  stylesheet.  If the network response of the stylesheet included an
+  `<code>Referrer-Policy</code>` header, the header's policy should take
+  precedence.
+
+  Both, the value of the referrer and the value of the referrer policy should be
```
estark37: unnecessary comma after "Both"

(I don't think I'm using this new github code review thing right)

index.src.html
estark37: Oh, now I see @domenic's comment outlining the different possibilities. Hmm.

estark37:
Ugh, `<script>` is a can of worms, I think. Suppose we say that, for example, requests initiated by a stylesheet use the stylesheet URL as the referrer and the stylesheet's Referrer-Policy header for the policy. I'm guessing we don't want `<script>`s to behave the same way, right? i.e. we wouldn't want to use the script URL as the referrer for requests that the script initiates, we'd use the document URL, which means we'd want to use the document referrer policy and not the Referrer-Policy header served with the script. It sucks that stylesheets and requests they initiate behave differently than scripts and requests they initiate.

FWIW I really like the simplicity of "always use the document/worker URL as the referrer and the document/worker's referrer policy, unless overridden for an individual request by a `referrerpolicy` attribute". But I guess it just doesn't make sense to use the document's referrer policy for stylesheets if the referrer is the stylesheet URL.

jeisinger:
ok, next attempt.

jeisinger:
btw, I think even if we add a referrerpolicy attribute to <link rel=stylesheet> at some point in the future, that referrer policy shouldn't influence the referrer policy for loading resources from that stylesheet.

index.src.html
```diff
@@ -655,6 +655,28 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+
+  For external stylesheets, the referrer policy should be
+  <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
+  `<code>Referrer-Policy</code>` header.
+
+  For inline stylesheets, the referrer policy is the containing {{Document}}'s
+  referrer policy.
+
+  Both the value of the referrer and the value of the referrer policy should be
+  captured at the time a given request is created.
```
domenic: Is this really the intention? Or is it at the time the stylesheet is created?

I think this only matters for inline style sheets where the containing document's referrer policy can change over time as meta elements are added/removed.

index.src.html
```diff
@@ -655,6 +655,28 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  stylesheets. However, implementations use the URL a given stylesheet was
+  loaded from as referrer for the requests for resources from that stylesheet.
+
+  Implementations should keep track of a referrer policy for each stylesheet
+  that should be used to create requests for resources from the respective
+  stylesheet. 
+
+  For external stylesheets, the referrer policy should be
+  <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
+  `<code>Referrer-Policy</code>` header.
```
domenic: Maybe explicitly state that the `referrerpolicy` attribute on link should not impact fetches from inside the stylesheet.

index.src.html
jeisinger: This is also when we capture the base URI which is used to resolve relative URLs, and yes, it should only matter for inline stylesheets.

index.src.html
jeisinger: sgtm

index.src.html
```diff
@@ -655,6 +655,32 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
 </section>
 
+<section class="informative">
+  <h2 id="integration-with-css">Integration with CSS</h2>
+
+  The CSS Standard does not specify how it fetches resources referenced from
+  stylesheets. However, implementations use the URL a given stylesheet was
+  loaded from as referrer for the requests for resources from that stylesheet.
+
+  Implementations should keep track of a referrer policy for each stylesheet
```
annevk: RFC2119 terms like "should" are reserved for normative requirements. We should keep them out of notes.

index.src.html
```diff
@@ -655,6 +655,32 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  that should be used to create requests for resources from the respective
```
annevk: Same here.

index.src.html
```diff
@@ -655,6 +655,32 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  stylesheet. 
+
+  For external stylesheets, the referrer policy should be
```
annevk: And here.

index.src.html
```diff
@@ -655,6 +655,32 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
+  `<code>Referrer-Policy</code>` header.
+
+  Note: If the stylesheet was loaded via a HTML <{link}> element with a
+  declared <{link/referrerpolicy}>, this referrer policy will not affect the
+  requests for resources referenced from the stylesheet.
+
+  For inline stylesheets, the referrer policy is the containing {{Document}}'s
+  referrer policy.
```
annevk: Maybe it's best to say here when it's captured? Or merge the next paragraph into this one.

jeisinger: merged the paragraphs

index.src.html
jeisinger: I removed the "style=informative" from the entire section, since it's normative until CSS got its stuff figured out.

jeisinger:
style="" should be handled like inline stylesheets (i.e. captured at parse time). Should I explicitly add that?

annevk:
Yeah, I think mentioning `style=""` would be good, since it doesn't end up in `document.styleSheets` so folks might expect it to work differently.

jeisinger:
done, ptal

annevk:
Looks reasonable to me now. Also paging @tabatkins since he said he might work on this at some point in CSS at which point we could remove this again.

index.src.html
```diff
@@ -656,6 +656,31 @@ <h2 id="integration-with-html">Integration with HTML</h2>
 </section>
 
 <section>
+  <h2 id="integration-with-css">Integration with CSS</h2>
+
+  The CSS Standard does not specify how it fetches resources referenced from
+  stylesheets. However, implementations use the URL a given stylesheet was
```
bzbarsky: Is this section informative or normative?  Because if this is meant to be normative, then "the URL a given stylesheet was loaded from" needs to be defined much more clearly than it is now.  If this is meant to be informative, the text is still confusing, of course...

index.src.html
```diff
@@ -656,6 +656,31 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  `<code>Referrer-Policy</code>` header.
+
+  Note: If the stylesheet was loaded via a HTML <{link}> element with a
+  declared <{link/referrerpolicy}>, this referrer policy will not affect the
+  requests for resources referenced from the stylesheet.
+
+  For inline stylesheets, and styles applied via an <code>style</code>
+  attribute on an element, the referrer policy is the containing {{Document}}'s
+  referrer policy.  Both the value of the referrer and the value of the
+  referrer policy should be captured at the time a given request is created.
```
bzbarsky: So not at stylesheet parse time, right?

What is expected to happen for directly-constructed stylesheets that are then filled with rules, once CSSOM supports that?

What is expected to happen if a single stylesheet (inline, or directly constructed) is attached to multiple documents?

index.src.html
```diff
@@ -656,6 +656,31 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+
+  For external stylesheets, the referrer policy should be
+  <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
+  `<code>Referrer-Policy</code>` header.
+
+  Note: If the stylesheet was loaded via a HTML <{link}> element with a
+  declared <{link/referrerpolicy}>, this referrer policy will not affect the
+  requests for resources referenced from the stylesheet.
+
+  For inline stylesheets, and styles applied via an <code>style</code>
```
bzbarsky: Oh, and s/an/a/ before "style".

bzbarsky:
And of course the future-looking CSSOM bits would best be dealt with by the CSS/CSSOM specs...

index.src.html
jeisinger: There's only one request for a sub-resource, even if it's used from multiple documents, so I'd say that whatever needs the resource first triggers the creation of the request, and therefore provides the referrer & referrer policy

index.src.html
bzbarsky: > we use "the document's URL" as referrer, without specifying this further.

No.  You use "the document's URL" with the "URL" bit being a link to the explicit definition of that term in the spec that defines it, and which can then be used to figure out what that URL actually is in any given circumstance.

> Would saying something like "the stylesheet's URL" address your concerns?

Not without a clear definition, either inline or via link to the relevant spec that does define it.

bzbarsky: Yes, and the spec that definition is in, plus other specs that might manipulate document URLs, defines what that URL is for each Document object.   For example, the very next paragraph says:

> Unless stated otherwise ... URL is "about:blank"

So in practice, an implementor has a field called "URL" on document objects.  When a document is created, set the field to "about:blank".  When some spec says to change the document's URL, change it accordingly.  For example, https://html.spec.whatwg.org/multipage/browsers.html#set-the-document%27s-address does this during navigation.  https://html.spec.whatwg.org/multipage/browsers.html#dom-history-pushstate step 10 does this when pushState or replaceState is called.  That sort of thing.

Then when some spec says to use the document's URL, use the value of that field.  Everything is clearly defined, no ambiguity.

jeisinger: yeah, i'm just whining a bit about the absence of all of this from CSS.

bzbarsky: Oh, I agree the CSS situation is rather sucky.  Can we just get @zcorpan or @tabatkins to just fix it already?  :(

If not, we could probably define it as https://drafts.csswg.org/cssom/#concept-css-style-sheet-location unless that's null or our load is not coming from a stylesheet, in which case it's the owner node's node document's URL (those would all be links to the relevant definitions) or the element's node document's URL, for inline stylesheets and style attributes respectively. And then we hope the CSS situation gets resolved sanely before there's a way to create stylesheets that have neither a location or an owner node...

(And I'm not even going to think about what happens if we ever get nontrivial ways to animate url values in CSS, for example taking part of the URL from one stylesheet and part from another.  Let's just assume people won't do that without thinking about the consequences.)

Issue #61, opened 2016-07-15
Travis config is broken


estark37:
Travis builds are spontaneously failing, e.g. https://travis-ci.org/w3c/webappsec-referrer-policy/builds/145066155

This is the same Travis config as CSP, so presumably @mikewest has the same problem, and maybe he knows how to fix it too?

Presumably we need to update setuptools but I'm not sure what the right way to do that is.

jdm:
Servo's also having ongoing pip troubles on travis.

mikewest:
I looked into this briefly earlier today; https://github.com/w3c/webappsec-secure-contexts/commit/2e7cc9778181cae07c0e2902258998e4e8cd35f6 will get you a little closer, but there's another failure with python and html5lib that I haven't had time to look into yet.

Perhaps @tabatkins can help you out in a PST time zone?

tabatkins:
Hmm, yesterday afternoon gsnedders said 

> gsnedders> plinss: FWIW, latest html5lib probably breaks some stuff, so probably want to make sure server doesn't upgrade

So it's possible there's something new in html5lib that I need to adjust Bikeshed for. I'll look into it.

PR #69, opened 2016-09-22
Fix travis ci config (#61)

jeisinger:
/cc @estark37 @mikewest


mikewest:
Does it work? If so, LGTM.

jeisinger:
it makes the travis job go further, but it still fails :-/

Issue #71, opened 2016-09-29
redact location.ancestorOrigins according to Referrer Policy


hillbrad:
@bzbarsky @dakami and I had a hallway discussion at the end of TPAC about the possibility of adding location.ancestorOrigins to Firefox.  bz has had longstanding concerns about the information this leaks to iframes.  We arrived at a local consensus that any leakage is roughly equivalent to what happens already with referrer, so it would make sense to redact ancestorOrigins according to referrer policy.  (and this could resolve that objection to a Mozilla implementation of ancestorOrigins)

wseltzer:
should this issue close along with #72?

jeisinger:
Closing this issue in favor of the html one

PR #72, opened 2016-09-29
redact ancestorOrigins according to referrer policy

hillbrad:
addresses #71


domenic:
This should be a pull request to the HTML standard, not a [monkey patch](https://annevankesteren.nl/2014/02/monkey-patch) in Referrer Policy. We just finished moving all the monkey patches out of referrer policy and should not introduce new ones.

hillbrad:
Sounds good, will do.

PR #73, opened 2016-10-01
Copy edit

uklauer:
Commit 01e48049982c48c8178202f34572e4c00dc8ab7b added a third policy to a list of two, but did not update the following sentence that referred back to “two policies”.


jeisinger:
hey, thx for the patch. Could you rebase it, I'd merge it then.

@wseltzer fyi... the ipr check fails, but it's small fix, so I think it's ok to accept it

uklauer:
Rebased on current master.

wseltzer:
ok

PR #78, opened 2016-10-18
Use "potentially trustworthy URLs"

hillbrad:
The concept of an "a priori authenticated URL" no longer exists in MIX, use "potentially trustworthy URLs" from SECURE CONTEXTS throughout.

Also fixes cases where an origin was being used as in input to determine if a URL was trustworthy.


index.src.html
```diff
@@ -108,9 +108,9 @@
       text: name; for: meta; url: attr-meta-name
       text: referrerpolicy; for: a; url: attr-a-referrerpolicy
       text: referrerpolicy; for: link; url: attr-link-referrerpolicy
-spec: MIX; urlPrefix: https://w3c.github.io/webappsec/specs/mixedcontent/
+spec: SECURE-CONTEXTS; urlPrefix: https://www.w3.org/TR/secure-contexts
```
domenic: Wrong URL, should be https://w3c.github.io/webappsec-secure-contexts/ ("TR stands for trash")

estark37:
Thanks, @hillbrad and @domenic! LGTM with @domenic's comment about the URL.

PR #79, opened 2016-10-31
Editorial: reduce dependency on anchors block

annevk:


annevk:
You're depending a lot on the bikeshed "anchors" block while most specifications don't need that. It would be much better to make use of the automatic linking so you know when definitions change or go out of-of-date.

This currently generates one bikeshed warning (though also removes some) as I have not marked one term in Fetch for export that you happen to need. That should be fixed by tomorrow.

annevk:
You might also want to take some time to see what more terms you can remove as the current setup is rather fragile.

estark37:
Ah, cool, I didn't know this was the right way to do things. Thanks! I get a bunch of errors from bikeshed when I make this, though:

```
$ make
bikeshed -f spec ./index.src.html
FATAL ERROR: No 'dfn' refs found for 'requests' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'client' that are marked for export.
FATAL ERROR: No 'idl' refs found for 'Request'.
FATAL ERROR: No 'dfn' refs found for 'origin' with for='request'.
FATAL ERROR: No 'dfn' refs found for 'current url' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'potentially trustworthy url'.
FATAL ERROR: No 'dfn' refs found for 'clients' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'request' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'referrer policy' with for='request'.
FATAL ERROR: No 'dfn' refs found for 'browsing contexts' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'responses' that are marked for export.
FATAL ERROR: No 'idl' refs found for 'Response'.
FATAL ERROR: No 'dfn' refs found for 'parse a header value'.
FATAL ERROR: No 'dfn' refs found for 'header list' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'response' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'referrer' with for='request'.
FATAL ERROR: No 'dfn' refs found for 'browsing context' that are marked for export.
FATAL ERROR: No 'dfn' refs found for 'local scheme' that are marked for export.
```

Any idea what's going on? Am I doing something wrong?

annevk:
@estark37 I suspect if you run "git pull" followed by "bikeshed update" in your local bikeshed repository those should go down a fair bit.

annevk:
I updated the PR to remove the last warning (and also remove a new warning due to changes in the URL Standard).

estark37:
Marked as non-substantive for IPR from ash-nazg.

estark37:
Ah, yes, that does the trick, thanks.

PR #80, opened 2016-10-31
Editorial: export terms for Fetch

annevk:


annevk:
Note that the generated resource is not ideal here. I tried to solve those problems in #79 so you might want to merge that first.

annevk:
This PR coupled with https://github.com/tabatkins/bikeshed/issues/855 will make it feasible for Fetch to no longer use an "anchors" block for Referrer Policy links.

annevk:
I'll rebase this.

annevk:
Done. By the way, why is there both a master and gh-pages branch?

estark37:
Thanks @annevk! The gh-pages branch makes the spec accessible at https://w3c.github.io/webappsec-referrer-policy/

annevk:
I understand, but wouldn't it be easier to just have a gh-pages branch then and do away with master?

estark37:
Not sure -- that's the way it's been set up since before my time, and seems to be the way other webappsec specs are set up too. @mikewest @jeisinger? Maybe it's sometimes convenient to be able to put changes on master without updating the editor's draft right away?

PR #83, opened 2016-12-21
Update "Integration with CSS" section to be more concrete

jeisinger:
@estark37 ptal

/cc @mikewest @annevk @bzbarsky

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -16,6 +16,12 @@ Repository: w3c/webappsec-referrer-policy
 </pre>
 
 <pre class="anchors">
+spec: CSSOM-1; urlPrefix: https://drafts.csswg.org/cssom-1/
+  type: dfn
+    text: CSS declaration block
+    text: location; url: concept-css-style-sheet-location
+    text: owner node; for: CSSStyleDeclaration; url: cssstyledeclaration-owner-node
+    text: owner node; for: CSSStyleSheet; url: concept-css-style-sheet-owner-node
 spec: FETCH; urlPrefix: https://fetch.spec.whatwg.org/
   type: dfn
     text: fetching
@@ -180,16 +186,16 @@ spec: html; type: element; text: a;
 
   <dl>
     <dt>
-      <a>referrer policy</a>
+      <a for="/">referrer policy</a>
     </dt>
     <dd>
-      A <a>referrer policy</a> modifies the algorithm used to populate the
+      A <a for="/">referrer policy</a> modifies the algorithm used to populate the
       <a><code>Referer</code></a> header when <a>fetching</a> subresources,
       prefetching, or performing navigations. This document defines the various
-      behaviors for each <a>referrer policy</a>.
+      behaviors for each <a for="/">referrer policy</a>.
 
       Every <a>environment settings object</a> has an algorithm for obtaining a
-      <a>referrer policy</a>, which is used by default for all <a for=/>requests</a>
+      <a for="/">referrer policy</a>, which is used by default for all <a for=/>requests</a>
       with that <a>environment settings object</a> as their <a for=request>client</a>.
     </dd>
 
@@ -233,7 +239,7 @@ spec: html; type: element; text: a;
     };
   </pre>
 
-  Each possible <a>referrer policy</a> is explained below. A detailed
+  Each possible <a for="/">referrer policy</a> is explained below. A detailed
   algorithm for evaluating their effect is given in the
   <a section href="#integration-with-fetch"></a> and
   <a section href="#algorithms"></a> sections.
@@ -466,8 +472,8 @@ spec: html; type: element; text: a;
 
   <h3 dfn export id="referrer-policy-empty-string">The empty string</h3>
 
-  The empty string "" corresponds to no <a>referrer policy</a>, causing a
-  fallback to a <a>referrer policy</a> defined elsewhere, or in the case where
+  The empty string "" corresponds to no <a for="/">referrer policy</a>, causing a
+  fallback to a <a for="/">referrer policy</a> defined elsewhere, or in the case where
   no such higher-level policy is available, defaulting to
   <a>"<code>no-referrer-when-downgrade</code>"</a>. This defaulting happens in
   the [[#determine-requests-referrer]] algorithm.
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -16,6 +16,15 @@
 </pre>
 
 <pre class="anchors">
+spec: CSSOM; urlPrefix: https://drafts.csswg.org/cssom/
```
domenic: It should not be necessary to add anything to anchors; Bikeshed autolinking should suffice for this, since CSSOM and DOM are already in Bikeshed.

index.src.html
```diff
@@ -635,25 +645,29 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
-  that should be used to create requests for resources from the respective
-  stylesheet.
-
-  For external stylesheets, the referrer policy should be
-  <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
-  `<code>Referrer-Policy</code>` header.
-
-  Note: If the stylesheet was loaded via a HTML <{link}> element with a
+  stylesheets. However, implementations should use the <a>location</a> of the
+  <a>stylesheet</a>, <a href="#strip-url">stripped for use as a referrer</a>,
```
mikewest: Nit: I don't think you need to strip the URL here. It gets stripped as part of https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer (step 4-5).

index.src.html
```diff
@@ -635,25 +645,29 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
-  Implementations should keep track of a referrer policy for each stylesheet
-  that should be used to create requests for resources from the respective
-  stylesheet.
-
-  For external stylesheets, the referrer policy should be
-  <a>"<code>no-referrer-when-downgrade</code>"</a> unless overwritten by an
-  `<code>Referrer-Policy</code>` header.
-
-  Note: If the stylesheet was loaded via a HTML <{link}> element with a
+  stylesheets. However, implementations should use the <a>location</a> of the
```
mikewest: Nit: `<a for="CSSStyleSheet">location</a>` (You're aiming for https://drafts.csswg.org/cssom/#concept-css-style-sheet-location, right?)

index.src.html
```diff
@@ -635,25 +645,29 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  <a>stylesheet</a>, <a href="#strip-url">stripped for use as a referrer</a>,
+  as referrer for the requests for resources from that <a>stylesheet</a>,
```
mikewest: Nit: `<a for="request">referrer</a>`, `<a>request</a>`

index.src.html
```diff
@@ -635,25 +645,29 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  request did not originate from a <a>stylesheet</a>, implementations should
+  use the <a>owner node</a>'s <a>node document</a>'s <a for="document">URL</a>,
+  <a href="#strip-url">stripped for use as a referrer</a>, or the
+  <a>element</a>'s <a>node document</a>'s <a for="document">URL</a>, <a
+  href="#strip-url">stripped for use as a referrer</a>, for <a>style
+  elements</a> and <a>style attributes</a> respectively. The referrer policy
+  should be the respective <a>document's referrer policy</a>.
+
+  Both the value of the referrer and the value of the referrer policy should be
+  captured at the time a given request is created.
```
mikewest: I think we can probably clean this up a bit by rearranging things. WDYT of something like:

> ```
> The CSS Standard does not specify how it fetches resources referenced from
> stylesheets. However, implementations should be sure to set the referrer-related
> properties of any <a>requests</a> initiated by stylesheets as follows:
>
> 1.   If a <a>CSS declaration block</a> is responsible for the request,
>      set the <a for="request">referrer</a> to the block's
>      <a for="CSSStyleDeclaration">owner node</a>'s
>      <a>node document</a>'s <a for="document">URL</a>, and the
>      <a for="request">referrer policy</a> to the block's
>      <a for="CSSStyleDeclaration">owner node</a>'s
>      <a>node document</a>'s <a for="Document">referrer policy</a>.
>
> 2.   If a <a>CSS style sheet</a> is responsible for the request,
>      and its <a for="CSSStyleSheet">location</a> is non-null,
>      set the <a for="request">referrer</a> to its
>      <a for="CSSStyleSheet">location</a>, and the <a for="request">referrer
>      policy</a> to its <a for="CSSStyleSheet">referrer policy</a>.
>
>      Issue: This requires that CSS style sheets process `Referrer-Policy`
>      headers, and store a <dfn for="CSSStyleSheet">referrer policy</a>
>      in the same way that <a for="Document" lt="referrer policy">`Document`s
>      do</a>.
>
> 3.  Otherwise, a <a>CSS style sheet</a> with a null location is responsible
>     is responsible for the request: set the <a for="request">referrer</a> to its
>     <a for="CSSStyleSheet">owner node</a>'s
>     <a>node document</a>'s <a for="document">URL</a>, and the
>     <a for="request">referrer policy</a> to the block's
>     <a for="CSSStyleSheet">owner node</a>'s
>     <a>node document</a>'s <a for="Document">referrer policy</a>.
>
> Note: Both the value of the <a>request</a>'s <a for="request">referrer</a> and
> <a for="request">referrer policy</a> are set based on the values at the time
> a given <a>request</a> is created. If a document's referrer policy changes
> during its lifetime, the policy associated with inline stylesheet requests will
> also change.
> ```

WDYT?

index.src.html
jeisinger: turns out that CSSOM doesn't export it's dfns, so I need to list them here :(

index.src.html
jeisinger: sgtm, thx

jeisinger:
ptal

index.src.html
```diff
@@ -635,25 +641,50 @@ <h2 id="integration-with-html">Integration with HTML</h2>
...
+  stylesheets. However, implementations should be sure to set the
+  referrer-related properties of any <a for="/">requests</a> initiated by stylesheets
+  as follows:
 
-  Implementations should keep track of a referrer policy for each stylesheet
-  that should be used to create requests for resources from the respective
-  stylesheet.
+  <ol>
+    <li>
+       If a <a>CSS declaration block</a> is responsible for the request,
```
mikewest: Tiny nit: If you enable markdown via `Markup Shorthands: markdown on`, you could write this as:

```
1.   Item 1.
2.   Item 2.
3.   Item 3.
```

Which is prettier! And easier to read! But has no normative impact at all so can easily be done in a future PR!

jeisinger: yeah, the rest of the spec also uses <ol> so I figured to stick with that for consistency

PR #92, opened 2017-01-11
Clarify referrer rules for inline styles

jeisinger:
@bzbarsky wdyt

### DIFF
--- a/index.html
+++ b/index.html
@@ -418,7 +418,7 @@
 	[data-algorithm]:not(.heading) {
 	  padding: .5em;
 	  border: thin solid #ddd; border-radius: .5em;
-	  margin: .5em 0;
+	  margin: .5em calc(-0.5em - 1px);
 	}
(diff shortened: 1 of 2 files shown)


bzbarsky:
This seems ok as far as it goes, but the claim that "Otherwise, a CSS declaration block that is part of an element's inline style properties is responsible for the request" is just not true for `<body background="whatever">` and similar...

jdm:
Can you ensure that web-platform-tests has tests covering this?

jeisinger:
Yes, I'll work on WPT tests before landing this in blink

jeisinger:
wpt is here: https://github.com/w3c/web-platform-tests/pull/5160

will polish the spec test to reflect this

jeisinger:
@bzbarsky ptal

tabatkins:
Looks ok to me with a fairly quick readthrough :+1:

PR #99, opened 2017-03-01
Editorial: use new extracting header list values hook from Fetch

### LINKED ISSUE #98: Fetch concept-header-parse reference is out of date
Looks like #concept-header-parse links should be updated to https://fetch.spec.whatwg.org/#extract-header-list-values (I think)

annevk:
Fixes #98.


estark37:
Marked as non-substantive for IPR from ash-nazg.

estark37:
Thanks!

PR #100, opened 2017-03-01
Editorial: reference origin terms from the HTML Standard

annevk:
RFC6454 is effectively obsolete. Also clean up some references to Fetch.


annevk:
This is best applied after #99. HTML might need to be generated again.

estark37:
Thanks @annevk! Regenerated HTML and merged

Issue #81, opened 2016-11-15
Add an example for multiple Referrer-Policy headers


estark37:
@marumari points out that the spec is a little confusing on how you can use the Referrer-Policy header to specify multiple policies. https://w3c.github.io/webappsec-referrer-policy/#unknown-policy-values could do with an example, showing that `Referrer-Policy: origin no-referrer` is not going to work but that `Referrer-Policy: origin, no-referrer` will.

marumari:
I know that _technically_ it's all listed deep within the HTTP specification, but for the average user I think it's unlikely that they'll be able to figure it out from simply reading the specification.  It doesn't help that similar specifications (such as HSTS and CSP) separate their directives with either spaces or semicolons.

justin-morgan:
+1, I've been banging my head against the wall on this for a while, and finally found the definition way down in the RFC (https://tools.ietf.org/html/rfc2616#section-4.2).  There is already an example section for unknown policy values, so it would be relatively easy to add an explicit HTTP response header implementation to match the example text.

annevk:
Probably good to have tests as well for edge cases such as:
```
Referrer-Policy: origin
Referrer-Policy: origin no-referrer
```

PR #101, opened 2017-03-06
Add an example for multiple Referrer-Policy headers

estark37:
This adds a couple examples for how to send multiple Referrer-Policy header
values in the section about falling back on unknown policy values.

Fixes https://github.com/w3c/webappsec-referrer-policy/issues/81

### DIFF
--- a/index.html
+++ b/index.html
@@ -1176,7 +1176,7 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version 0060add2b6463a7a7a099173ff2c8a54d3e0ce53" name="generator">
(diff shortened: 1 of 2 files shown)


estark37:
@jeisinger wdyt?

domenic:
(Random bystander question, hopefully helpful.)

Are these two styles exactly equivalent? If so stating that explicitly might be good. E.g. I could imagine a world where things were more like CSS and the comma-separated version just did nothing if any of the values were unsupported.

estark37:
Thanks @domenic, they are equivalent and I added that into the text.

PR #102, opened 2017-03-07
Relax ABNF for invalid policy tokens

estark37:
@annevk pointed out in https://github.com/w3c/web-platform-tests/pull/5054#issuecomment-284664280 that when `policy-token` is defined as a collection of literals corresponding to the valid referrer policies, the browser should technically completely disregard a header of the form `Referrer-Policy: origin, blah`. Instead, we want the browser to ignore `blah` as an unknown policy value and fall back to `origin`.

So this PR relaxes the ABNF of the `Referrer-Policy` header to parse invalid policy values.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -22,6 +22,9 @@ spec: CSSOM-1; urlPrefix: https://drafts.csswg.org/cssom-1/
     text: location; url: concept-css-style-sheet-location
     text: owner node; for: CSSStyleDeclaration; url: cssstyledeclaration-owner-node
     text: owner node; for: CSSStyleSheet; url: concept-css-style-sheet-owner-node
+spec: Fetch; urlPrefix: https://fetch.spec.whatwg.org/
+  type: dfn
+    text: ABNF; url: abnf
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     urlPrefix: semantics.html
@@ -97,7 +100,6 @@ spec: WSC-UI; urlPrefix: http://www.w3.org/TR/2010/REC-wsc-ui-20100812/
     text: TLS-protected; url: typesoftls
 spec: RFC5234; urlPrefix: https://tools.ietf.org/html/rfc5234
   type: dfn
-    text: ABNF; url:
     text: ALPHA; url: appendix-B.1
 spec: RFC7231; urlPrefix: https://tools.ietf.org/html/rfc7231
   type: dfn
--- a/index.src.html
+++ b/index.src.html
@@ -22,9 +22,6 @@ spec: CSSOM-1; urlPrefix: https://drafts.csswg.org/cssom-1/
     text: location; url: concept-css-style-sheet-location
     text: owner node; for: CSSStyleDeclaration; url: cssstyledeclaration-owner-node
     text: owner node; for: CSSStyleSheet; url: concept-css-style-sheet-owner-node
-spec: Fetch; urlPrefix: https://fetch.spec.whatwg.org/
-  type: dfn
-    text: ABNF; url: abnf
 spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
   type: dfn
     urlPrefix: semantics.html
@@ -99,7 +96,7 @@ spec: WSC-UI; urlPrefix: http://www.w3.org/TR/2010/REC-wsc-ui-20100812/
   type: dfn
     text: TLS-protected; url: typesoftls
 spec: RFC5234; urlPrefix: https://tools.ietf.org/html/rfc5234
-  type: dfn
+  type: grammar
     text: ALPHA; url: appendix-B.1
 spec: RFC7231; urlPrefix: https://tools.ietf.org/html/rfc7231
   type: dfn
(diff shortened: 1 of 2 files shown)


estark37:
@annevk does this address the issue you raised in https://github.com/w3c/web-platform-tests/pull/5054#issuecomment-284664280?

I'm not sure if any changes are needed to https://w3c.github.io/webappsec-referrer-policy/#parse-referrer-policy-from-header. I think it works as is, by throwing out any `token` which is not a referrer policy?

annevk:
So this works, but this also means that a value such as `origin no-referrer` results in failure since you don't allow spaces. That seems fine by the way, just pointing it out since it will affect the outcome of one of the tests.

The parsing steps still look good indeed.

@mikewest might also want to look at this since I suspect CSP might suffer from similar issues, but not sure.

annevk:
Oh, for the term ABNF you might want to reference https://fetch.spec.whatwg.org/#abnf or say something equivalent.

mikewest:
I'm pretty sure the structure of directives in CSP is loose enough to formally allow most everything: https://w3c.github.io/webappsec-csp/#framework-directives. The ABNF for parsing source lists probably needs an "everything else" bucket, but I'm not sure it's a good idea to drop the list of valid keywords as this patch does. I find it pretty useful to be able to link to things like https://w3c.github.io/webappsec-csp/#grammardef-unsafe-inline (if only because it helps me catch spelling errors).

annevk:
An alternative that @estark37 might be willing to consider is leaving the explicit keywords in and defining something like an "extension-token" production that is ALPHA / "-". HTTP uses that at times I believe.

mikewest:
Right, that's what I was clumsily suggesting with an "everything else" bucket. Something like:

```
"Referrer-Policy:" 1#(policy-token / extension-token)

policy-token    = "no-referrer" / "no-referrer-when-downgrade" / "strict-origin" / "strict-origin-when-cross-origin" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
extension-token = ALPHA / "-"
```

estark37:
Ohh ok, perfect. Went back to the explicit keywords with an `extension-token` bucket. Thanks!

index.src.html
```diff
@@ -519,17 +525,23 @@ <h3 id="referrer-policy-header">Delivery via Referrer-Policy header</h3>
...
-    <dfn export>policy-token</dfn>   = "no-referrer" / "no-referrer-when-downgrade" / "strict-origin" / "strict-origin-when-cross-origin" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
+    <dfn export>policy-token</dfn> = "no-referrer" / "no-referrer-when-downgrade" / "strict-origin" / "strict-origin-when-cross-origin" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
+    <dfn>extension-token</dfn> = 1*( <a>ALPHA</a> / "-" )
   </pre>
 
   Note: The header name does not share the HTTP Referer header's misspelling.
 
+  Note: The purpose of <a>extension-token</a> is so that browsers do not fail to
+  parse the entire header field if it includes an unknown policy
+  value. [[#unknown-policy-values]] describes in greater detail how new policy
```
mikewest: Nit: Line wrapping is a bit weird here.

index.src.html
```diff
@@ -519,17 +525,23 @@ <h3 id="referrer-policy-header">Delivery via Referrer-Policy header</h3>
...
-  following ABNF grammar:
+  following <a>ABNF</a> grammar:
 
   <pre>
-    "Referrer-Policy:" 1#<a>policy-token</a>
+    "Referrer-Policy:" 1#(<a>policy-token</a> / <a>extension-token</a>)
   </pre>
   <pre>
-    <dfn export>policy-token</dfn>   = "no-referrer" / "no-referrer-when-downgrade" / "strict-origin" / "strict-origin-when-cross-origin" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
+    <dfn export>policy-token</dfn> = "no-referrer" / "no-referrer-when-downgrade" / "strict-origin" / "strict-origin-when-cross-origin" / "same-origin" / "origin" / "origin-when-cross-origin" / "unsafe-url"
```
mikewest: Nit: This should probably be defined as a `grammar` type. The simplest way to do that is to add `dfn-type="grammar"` to the `<pre>` wrapping the ABNF bits.

index.src.html
```diff
@@ -22,6 +22,9 @@
     text: location; url: concept-css-style-sheet-location
     text: owner node; for: CSSStyleDeclaration; url: cssstyledeclaration-owner-node
     text: owner node; for: CSSStyleSheet; url: concept-css-style-sheet-owner-node
+spec: Fetch; urlPrefix: https://fetch.spec.whatwg.org/
+  type: dfn
+    text: ABNF; url: abnf
```
mikewest: I don't think this is a good idea. Fetch doesn't define ABNF, RFC5234 does. Instead, I'd suggest that you add something like the text from https://github.com/w3c/webappsec-csp/commit/40d2e607855b91f91cfb0f1a183be1b3ce209553 which explicitly points to RFC7230's `#rule` definition, which is all that you're using here.

@annevk: Perhaps this is something we could move to INFRA?

annevk: Either Fetch or HTTP seems like the right place since you're effectively building on those anyway. INFRA is for when there's nothing else to build on.

estark37: Done

index.src.html
estark37: Done, thanks

index.src.html
estark37: Fixed

estark37:
@mikewest does the latest iteration lgty?

mikewest:
Yes! LGTM. Sorry, I thought you'd landed this already. :)

PR #103, opened 2017-04-20
Add a link to web-platform-tests to the top of the spec

foolip:
This is similar to https://github.com/w3c/webappsec-csp/pull/206


estark37:
Thanks @foolip!

PR #107, opened 2017-10-14
Fix HTML link: creation URL

domfarolino:
In 8.3 > step 3 > "client" case > step 2 there is an invalid link to HTML definition of a creation URL (now existing at #concept-environment-creation-url) which this PR addresses.

Also, not sure why the generated index file has soooo many changes, I only touched the src file and ran bikeshed. Is this because bikeshed has been updated since the last commit to this spec or did I mess something up?

***
[Preview](https://s3.amazonaws.com/pr-preview/domfarolino/webappsec-referrer-policy/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/webappsec-referrer-policy/8232912...domfarolino:65d5c52.html)

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -77,7 +77,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: environment settings object
       text: responsible browsing context
       text: global object; for: environment settings object; url: concept-settings-object-global
-      text: creation URL
+      text: creation URL; url: concept-environment-creation-url
     urlPrefix: workers.html
       text: running a worker; url: run-a-worker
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -77,7 +77,7 @@
       text: environment settings object
       text: responsible browsing context
       text: global object; for: environment settings object; url: concept-settings-object-global
-      text: creation URL
+      text: creation URL; url: concept-environment-creation-url
```
annevk: If you use `<a for=environment>creation URL</a>` at the point of usage this should not be needed at all.

PR #110, opened 2017-11-07
Linked testing policy and fixed a few links

plehegar:
per
https://www.w3.org/2017/11/06-webappsec-minutes.html#item16


mikewest:
:+1:

PR #122, opened 2019-06-05
Limit `referer` header's value to 4k.

mikewest:
If the `referer` header's value is a URL with a length of more than 4k, strip it down to an origin.

Fixes https://github.com/whatwg/fetch/issues/903.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/pull/122.html" title="Last updated on Oct 21, 2019, 10:21 AM UTC (039ee88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/122/1a1f47b...039ee88.html" title="Last updated on Oct 21, 2019, 10:21 AM UTC (039ee88)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -812,6 +812,10 @@ spec: html; type: element; text: a;
       <var>referrerSource</var> for use as a referrer.</a>
     </li>
     <li>
+      If <var>referrerURL</var>'s length is greater than 4096, set <var>referrerURL</var>
+      to <var>referrerOrigin</var>.
+    </li>
+    <li>
       Let <var>referrerOrigin</var> be the result of
       <a href="#strip-url">stripping <var>referrerSource</var> for use as a
       referrer</a>, with the <code><a>origin-only flag</a></code> set to
--- a/index.src.html
+++ b/index.src.html
@@ -812,16 +812,16 @@ spec: html; type: element; text: a;
       <var>referrerSource</var> for use as a referrer.</a>
     </li>
     <li>
-      If the result of <a lt="url serializer">serializing</a> <var>referrerURL</var> is a string
-      whose length is greater than 4096, set <var>referrerURL</var> to <var>referrerOrigin</var>.
-    </li>
-    <li>
       Let <var>referrerOrigin</var> be the result of
       <a href="#strip-url">stripping <var>referrerSource</var> for use as a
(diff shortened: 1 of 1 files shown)


arturjanc:
I <3 this change! 

Is 4k restrictive enough to prevent the request from being too large and force server errors (especially when combined with other parts of the request that can be controlled by a cross-origin attacker, e.g. the URL)? As in, do we know the maximum request sizes supported by major HTTP servers?

mikewest:
1. 4k is more restrictive than nothing. :)
2. My recollection is that folks were pushing ~16k to cause errors. I'm sure some server somewhere will explode with 4k, because the internet, but this seems like a substantial improvement that's pretty safe to ship, as it affects only ~0.01% of requests. 2k affects an order of magnitude more... We can probably make that work if we need to, but I'm happier starting with a limit we can ~comfortably ship.

mikewest:
(Also, @dveditz suggested that we might also want to limit overall URL length, perhaps by capping `pushState()` in some way. I don't have time right now to dig into that, but it's another approach that might be reasonable to combine with this one!)

index.src.html
```diff
@@ -811,6 +811,10 @@ <h3 id="determine-requests-referrer" dfn export>
       Let <var>referrerURL</var> be the result of <a href="#strip-url">stripping
       <var>referrerSource</var> for use as a referrer.</a>
     </li>
+    <li>
+      If the result of <a lt="url serializer">serializing</a> <var>referrerURL</var> is a string
+      whose length is greater than 4096, set <var>referrerURL</var> to <var>referrerOrigin</var>.
+    </li>
```
annevk: You need to move this one step down since referrerOrigin isn't initialized yet. Also, you can use `<a for=string>length</a>` (= https://infra.spec.whatwg.org/#string-length) these days.

mikewest: Ugh. I should have followed up on this months ago. I addressed your feedback in https://github.com/w3c/webappsec-referrer-policy/pull/122/commits/906e3b868c61056c31d6a4d76bdf8adcd101164f a while back, and rebased today in https://github.com/w3c/webappsec-referrer-policy/pull/122/commits/039ee88ab9aa130f97ce14c87a6e7145247ff6aa.

Still ok with the verbage? If so, I'll land it.

PR #124, opened 2019-10-15
Make 'unsafe-url' an alias for 'no-referrer-when-downgrade'

mikewest:
This patch converts 'unsafe-url' to be an alias for 'no-referrer-when-downgrade'.
I think we're past the point at which 'unsafe-url' is a useful safety valve for
folks migrating to HTTPS, and it seems reasonable to remove this footgun.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/pull/124.html" title="Last updated on Feb 16, 2021, 11:23 PM UTC (607ea88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/124/20aad3a...607ea88.html" title="Last updated on Feb 16, 2021, 11:23 PM UTC (607ea88)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/pull/124.html" title="Last updated on May 24, 2022, 6:31 PM UTC (607ea88)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/124/20aad3a...607ea88.html" title="Last updated on May 24, 2022, 6:31 PM UTC (607ea88)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1029,7 +1029,7 @@ Possible extra rowspan handling
 		}
 	/* } */
 
-	@supports (display:grid) {
+	@supports (display:grid) and (display:contents) {
 		/* Use #toc over .toc to override non-@supports rules. */
 		#toc {
 			display: grid;
@@ -1212,9 +1212,9 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version 018be3f805d3272ce699a232398d6e3902d1ea8d" name="generator">
+  <meta content="Bikeshed version 08c4b0e94d147852f66673459784d3429bb3bda1" name="generator">
   <link href="http://www.w3.org/TR/referrer-policy/" rel="canonical">
-  <meta content="cd3894ac6bdc097abed1b8584d8e96bb0fca00ec" name="document-revision">
+  <meta content="20aad3a785a5ba2ab67c0780c9cded9c45e37735" name="document-revision">
 <style>/* style-md-lists */
 
 /* This is a weird hack for me not yet following the commonmark spec
@@ -1461,7 +1461,7 @@ c-[il] { color: #000000 } /* Literal.Number.Integer.Long */
   <div class="head">
    <p data-fill-with="logo"><a class="logo" href="https://www.w3.org/"> <img alt="W3C" height="48" src="https://www.w3.org/StyleSheets/TR/2016/logos/W3C" width="72"> </a> </p>
    <h1>Referrer Policy</h1>
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2018-10-25">25 October 2018</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft, <time class="dt-updated" datetime="2019-10-15">15 October 2019</time></span></h2>
    <div data-fill-with="spec-metadata">
     <dl>
      <dt>This version:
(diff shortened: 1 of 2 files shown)


mikewest:
cc @annevk @johnwilander: Any objections to this from Mozilla or Apple folks?

mikewest:
(Note that this also has the effect of turning `<meta name="referrer" value="always">` into `<meta name="referrer" value="default">`.)

domenic:
You may want to update https://html.spec.whatwg.org/multipage/semantics.html#meta-referrer as well. I think doing so is technically redundant, as it'll just flow through the system as unsafe-url until it reaches the algorithm you're patching here, but it seems nicer.

mikewest:
@domenic: Yes, that was the comment in https://github.com/w3c/webappsec-referrer-policy/pull/124#issuecomment-542061403. I'm happy to update HTML directly since it would make things more clear for folks who only read that document.

annevk:
`unsafe-url` was added because otherwise search (and other) providers would have annoying non-HTTPS redirectors. What would prevent that from happening again? No browser offering this?

Would it be an option to wait with landing this until it's successfully shipped in one browser?

mikewest:
Baidu is the only search engine I know of that uses `<meta content="always" name="referrer">`. They might introduce such a bouncer, which would be unfortunate.

Other entities that are using referrer policy seem to be doing so in a way more consistent with its original intent: given search engines the ability to send something _less_ than the whole URL so that query strings didn't leak. That's the way Google, Facebook, Twitter, etc use the mechanism, and that's become easier over time given the adoption of HTTPS we've seen over time. My assertion at the moment is that we're over that line.

I'm happy to wait on landing these until we figure out if we can ship it. That's not a problem at all from my perspective.

mikewest:
cc @EricLaw1979, who I should have CC'd earlier.

ericlaw1979:
I'm very much in favor of getting rid of everything related to HTTP, but this one feels like a bit of stretch to me.

While it's a footgun, it's a clearly labelled footgun. Is there any evidence that any site explicitly requesting unsafe behavior is surprised to be getting what it asked for?

mikewest:
@ericlaw1979: The feet I'm worried about aren't the developers', but the users'. Developers can take their time, and be _really_ sure that they're aiming between their users' toes and not at them, but that's a risk against which the users' agent should intercede.

For clarity: I'd like you to consider this patch to be baby-step 1 towards removing referrer information from non-secure pages entirely, which https://github.com/w3c/webappsec-referrer-policy/pull/126 would do. Stepping towards that goal seems reasonable, but if you think it would be more effective to communicate it directly, I'm happy to try. :)

mikewest:
(An alternate babyish step would be to drop this patch, and instead strip all non-secure requests' `referer` headers to at most an origin, regardless of a page's declared policy. I'd be happy with that option too.)

annevk:
@domfarolino @krgovind what's the status on this?

krgovind:
Is the next step for browsers to attempt shipping this feature?

My apologies. For Chrome, I think this got dropped between @mikewest 's team and mine. :)

We'll re-group internally for next steps.

mikewest:
I slapped together a Chromium telemetry patch in https://chromium-review.googlesource.com/c/chromium/src/+/2589875.

While our current telemetry shows that [~60.14% of page views set a referrer policy](https://chromestatus.com/metrics/feature/timeline/popularity/593) the upper bar for the impact of this change is likely to be significantly lower, as the majority of usage seems concentrated in a few sites that use restrictive policies. Based on a quick skim of HTTP Archive (query below), only ~0.02% of pages in the dataset are delivered over HTTPS, set a referrer policy, and contain `unsafe-url` in their payload (compared to ~0.35% total pages delivered over HTTPS and setting a referrer policy). That's a very loose measurement of course, as scripts might do the work on the page's behalf, but it's a good directional signal. If the use counter added above shows similar data, I'd be happy to push on this a bit. How do you feel about it a year later, @ericlaw1979? 

```sql
SELECT
  COUNT(*) * 100.0 / (SELECT COUNT(*) FROM `httparchive.pages.2020_10_01_desktop`) AS percent
FROM
  `httparchive.pages.2020_10_01_desktop`
WHERE
  STARTS_WITH(url, "https://")
  AND JSON_EXTRACT(payload, '$._blinkFeatureFirstUsed.Features.243') IS NOT NULL
  AND STRPOS(LOWER(payload), "unsafe-url") != 0
```

ericlaw1979:
Non-secure HTTP needs to die, die, die, and I'm still in favor of work toward that end. 

Having said that, I'm still worried about this change and particularly its impact on non-Internet software (decades of Enterprise intranet apps). And because it's enterprise, telemetry may be limited. If we take away this token, we leave vendors without a straightforward way to unblock update deployments.

We're _still_ finding even public sites broken by the `strict-origin-when-cross-origin` change (crbug.com/1112375) and because of the failure mode, it often takes a while to even figure out what caused regressions. 

In Chrome today, `no-referrer-when-downgrade` still strips the referrer when navigating to a secure HTTP context (http://localhost). While such navigations are presumably uncommon, this still feels like a weird gap?

krgovind:
> In Chrome today, `no-referrer-when-downgrade` still strips the referrer when navigating to a secure HTTP context (http://localhost). While such navigations are presumably uncommon, this still feels like a weird gap?

That's interesting. Is that a bug in Chrome's implementation, because [the spec](https://www.w3.org/TR/referrer-policy/#referrer-policy-no-referrer-when-downgrade) appears to allow navigations to potentially trustworthy URLs (which includes localhost) to get the full URL.

domfarolino:
While Chrome thinks this PR is indeed headed in the right direction, this has fallen off of our priority list. Since nobody has the bandwidth right now to ship this, and we haven't heard in this thread that this is a priority for any other browser right now, I think it makes sense to close this PR. I'll wait a few days and close this out unless we hear from anyone.

If we want to revisit this once it becomes more likely browsers could ship it, we can easily re-open this PR or recreate it.

Issue #121, opened 2019-05-28
What default policy should new features use?


jeremyroman:
This came up in WICG/portals#136. We've generally aimed for portals to lean on the side of exposing less information, but it's also desirable for its behavior to be consistent with how other features behave.

AFAICT, iframe and other subresource fetches all default to the document's referrer policy (which is, by default, `"no-referrer-when-downgrade"` per https://fetch.spec.whatwg.org/#main-fetch.

Would it be recommended for new features to rely on the existing logic (and authors to override the default if they want `"no-referrer"`, `"same-origin"` or similar), or should extra plumbing be added so that new features can have a different global default referrer policy?

@domenic @lucasgadani

annevk:
I think we should allow for more restrictive defaults or have a more restrictive default going forward that new features get to use.

fmarier:
Firefox has been using a [default of `strict-origin-when-cross-origin` in private browsing](https://bugzilla.mozilla.org/show_bug.cgi?id=587523) and this could be a nicer default for new features.

domfarolino:
I think that this can be closed since we've updated the default to strict-origin-when-cross-origin, but Jeremy let me know if there's more discussion you'd like to have.

PR #125, opened 2019-10-15
Default to 'strict-origin-when-cross-origin'.

mikewest:
This addresses w3c/webappsec-referrer-policy#121, and will be followed
with a corresponding patch to Fetch (where the default is currently
defined).

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -283,8 +283,6 @@ spec: html; type: element; text: a;
     <a><code>Referer</code></a> header.
   </div>
 
-  This is a user agent's default behavior, if no policy is otherwise specified.
-
   <h3 dfn export id="referrer-policy-same-origin">"<code>same-origin</code>"</h3>
 
   The <a>"<code>same-origin</code>"</a> policy specifies that a
@@ -443,6 +441,8 @@ spec: html; type: element; text: a;
     <a><code>Referer</code></a> header.
   </div>
 
+  This is a user agent's default behavior, if no policy is otherwise specified.
+
   <h3 dfn export id="referrer-policy-unsafe-url" oldids="referrer-policy-state-unsafe-url">"<code>unsafe-url</code>"</h3>
 
   The <a>"<code>unsafe-url</code>"</a> policy specifies that a full URL,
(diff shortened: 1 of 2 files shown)


mikewest:
cc @annevk and @johnwilander again for the same question.

domenic:
You may want to change the legacy `<meta>`-only mapping in HTML too: https://html.spec.whatwg.org/multipage/semantics.html#meta-referrer

annevk:
Would it be an option to wait with landing this until it's successfully shipped in one browser?

index.src.html
```diff
@@ -467,9 +470,8 @@ <h3 dfn export id="referrer-policy-empty-string">The empty string</h3>
 
   The empty string "" corresponds to no <a for="/">referrer policy</a>, causing a
   fallback to a <a for="/">referrer policy</a> defined elsewhere, or in the case where
-  no such higher-level policy is available, defaulting to
-  <a>"<code>no-referrer-when-downgrade</code>"</a>. This defaulting happens in
-  the [[#determine-requests-referrer]] algorithm.
+  no such higher-level policy is available, falling back to the <a>default referrer
+  policy</a>. This happens in Fetch's main fetch algorithm, for example.
```
annevk: Where else would it happen?

ehsan:
> Would it be an option to wait with landing this until it's successfully shipped in one browser?

FWIW [Firefox since version 59](https://blog.mozilla.org/security/2018/01/31/preventing-data-leaks-by-stripping-path-information-in-http-referrers/) has been shipping this as the default referrer policy in private browsing mode.

annevk:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1589074.

kiding:
Safari seems to do a similar thing by default [since 13.0.4/13.3](https://webkit.org/blog/9661/preventing-tracking-prevention-tracking/); but only for third-party requests, not for cross-origin _top navigations._

johnwilander:
Safari’s change is not a change of default setting. Sites can’t change it, only users can. Third-party referrers are downgraded to origin by default regardless of the referee policy set by the website.

index.src.html
krgovind: @mikewest I'm working on shepherding this change through the specs. Could you help answer @annevk 's question? I'm not sure if this is purely a clarification question or if it needs to result in a change in this PR?

krgovind:
@mikewest Would it make sense to abandon this PR in favor of #142?

alshamiri5 (commented):
ok

Issue #123, opened 2019-09-16
Inconsistencies with "same-origin" requests


domfarolino:
The [origin-when-cross-origin](https://www.w3.org/TR/referrer-policy/#referrer-policy-origin-when-cross-origin) section (and its "strict-" complement) mentions:

> and only the ASCII serialization of the origin of the request client is sent as referrer information when making cross-origin requests from a particular client.

However, the [reference](https://www.w3.org/TR/referrer-policy/#ref-for-referrer-policy-origin-when-cross-origin-5) to "origin-when-cross-origin" (and "strict-") from "determine a request's referrer" returns the origin of the request's [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer), which I believe is not always the same as request's client's origin. I believe they are often the same, however when loading module scripts, the module script baseURL is used as the referrer for its descendants (see [this](https://html.spec.whatwg.org/multipage/webappapis.html#fetching-scripts:internal-module-script-graph-fetching-procedure-2)) even though the descendant and parent share the same fetch client settings object. In this case _I believe_ "referrerOrigin" and request's client's origin are not the same, so the above block quote is misleading. Depending on the intentions of the spec, I see two ways forward:

1. Update the above quote to mention the origin of the referrer will be sent as the referrer for cross-origin requests etc, instead of the request client's origin
2. Modify "determine a request's referrer" to return the request's client's origin instead of `referrerOrigin` in these cases

/cc @yutakahirano @domenic (due to module script involvement)

### Edit:
FWIW going with option (1) will align the spec with the behavior we are trying to achieve with Chrome, and the behavior Firefox currently has (see [this test](https://wpt.fyi/results/html/semantics/scripting-1/the-script-element/module/referrer-origin-when-cross-origin.sub.html?label=experimental&label=master&aligned) & [this test](https://wpt.fyi/results/html/semantics/scripting-1/the-script-element/module/referrer-same-origin.sub.html?label=experimental&label=master&aligned)).

yutakahirano:
@mikewest @nyaxt @estark37

mikewest:
It makes sense to me that we'd strip the request's referrer down to an origin, and I expect that the current text just didn't take module scripts' referrer behavior into account. Have other browsers aligned on the same behavior here? If so, we should lock it in with some tests, and change the spec accordingly.

jeisinger:
[...]

PR #129, opened 2019-12-02
Stop using environment and request's origin in referrer calculation

yutakahirano:
***
<a href="https://pr-preview.s3.amazonaws.com/yutakahirano/webappsec-referrer-policy/pull/129.html" title="Last updated on Dec 2, 2019, 9:40 AM UTC (d455238)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/129/048f723...yutakahirano:d455238.html" title="Last updated on Dec 2, 2019, 9:40 AM UTC (d455238)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -845,16 +845,13 @@ spec: html; type: element; text: a;
         <dd>
           <ol>
             <li>
-              If |environment| is not null:
-              <ol>
-                <li>
-                  If |environment|'s <a for="environment settings object">HTTPS state</a> is "`modern`" <em>and</em>
-                  <var>request</var>'s
-                  <a href="https://fetch.spec.whatwg.org/#concept-request-current-url">current
-                  URL</a> is not a <a>potentially trustworthy
-                  URL</a>, then return <code>no referrer</code>.
-                </li>
-              </ol>
+              If |referrerURL| is <code>no referrer</code>, then return <code>no referrer</code>.
+            </li>
+            <li>
+              If |referrerURL| is a <a>potentially trustworthy URL</a> and <var>request</var>'s
+              <a href="https://fetch.spec.whatwg.org/#concept-request-current-url">current
+                URL</a> is not a <a>potentially trustworthy URL</a>, then return
+              <code>no referrer</code>.
             </li>
             <li>Return |referrerOrigin|.
           </ol>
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -864,22 +861,18 @@ <h3 id="determine-requests-referrer" dfn export>
...
-                  <a href="https://fetch.spec.whatwg.org/#concept-request-current-url">current
-                  URL</a> is not a <a>potentially trustworthy URL</a>
-                  <ol>
-                      <li>Return <code>no referrer</code>.
-                  </ol>
-                </li>
-              </ol>
+              If the  <a for=url>origin</a> of |referrerURL| and the <a for=url>origin</a> of
+              <var>request</var>'s <a for=request>current url</a> are <a lt="same origin">the
+              same</a>, then return |referrerURL|.
```
domfarolino: These look good, but I wonder if it would be cleaner to pull these comparisons out to a variable `same origin with referrer` and refer to it, since we do the comparisons several times.

domfarolino:
This will also require a Firefox bug, since I think their implementation will need to change /cc @annevk just FYI.

I can take care of updating the WPTs if this gets approved.

yutakahirano:
Yeah, I'd like to hear Firefox's (and Safari's) opinions before landing this change, probably at #123.

index.src.html
jeisinger: yes, also looks good to me.

index.src.html
```diff
@@ -845,16 +845,13 @@ <h3 id="determine-requests-referrer" dfn export>
...
-              <ol>
-                <li>
-                  If |environment|'s <a for="environment settings object">HTTPS state</a> is "`modern`" <em>and</em>
-                  <var>request</var>'s
-                  <a href="https://fetch.spec.whatwg.org/#concept-request-current-url">current
-                  URL</a> is not a <a>potentially trustworthy
-                  URL</a>, then return <code>no referrer</code>.
-                </li>
-              </ol>
+              If |referrerURL| is <code>no referrer</code>, then return <code>no referrer</code>.
```
domfarolino: I don't think instances of this are necessary, since the spec indicates that Fetch will not call into this algorithm if the referrer string is `no referrer`. Do you agree?

domfarolino:
Closing this in favor of https://github.com/w3c/webappsec-referrer-policy/pull/135

PR #135, opened 2020-05-12
Use request's referrer for more things, instead of request's client

### LINKED ISSUE #123: Inconsistencies with "same-origin" requests

domfarolino:
This PR fixes #123, and is meant to take over #129, which the author (@yutakahirano) does not have time to work on. This PR addresses my review of #129, as well as takes care of other things. Specifically, in total:

 - Stops comparing the origins of a request's [client](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-origin) and request's [current URL](https://fetch.spec.whatwg.org/#concept-request-current-url) when determining if a request is same-origin or not
    - Instead we compare the origins of a request's resolved referrer (its referrer string after "client" has been resolved") and its current URL
    - This fixes #123, and is tested
 - For the `strict-*` referrer policies, stop comparing request's client's [HTTPS state](https://html.spec.whatwg.org/multipage/webappapis.html#https-state) to request's currentURL's [trustworthiness](https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-url)
    - Instead compare a request's resolved referrer's potentially-trustworthiness against its current URL is potentially-trustworthiness
    - We should probably have a test for the case where an HTTP document sets a `strict-*` referrer policy. It fetches an HTTPS module script, which then fetches an HTTP descendant script. The request for the descendant should be sent with no referrer, since the referrer string is potentially-trustworthy, but the destination URL was not.
    - Test has been written in https://github.com/web-platform-tests/wpt/pull/23552
    - @jeisinger LGTM'd this change as part of https://github.com/w3c/webappsec-referrer-policy/pull/129#discussion_r353593951
 - Adds more examples demonstrating the effect that a referring resource such as a module script can have on something like a descendant script
 - Introduces the term "resolved referrer", which just means a request's referrer, after the "client" value has been resolved. This definition is a little hacky, so I welcome opinions. It just points to the `|referrerSource|` variable from the _determine request's referrer_ algorithm
 - Changes the definitions `{same,cross}-origin request` => `{same,cross}-origin-referrer request`, which more accurately describes that the referrer policies that perform same-origin comparisons, care specifically about whether a request is same-origin _relative to its referrer_, not whether the request is same-origin relative to its client (these can be different, as per #123)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/pull/135.html" title="Last updated on Jun 5, 2020, 12:14 AM UTC (68c94b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/135/fb230a1...68c94b1.html" title="Last updated on Jun 5, 2020, 12:14 AM UTC (68c94b1)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -112,6 +112,7 @@ spec: RFC2616; urlPrefix: https://tools.ietf.org/html/rfc7231
 <pre class="link-defaults">
 spec: html; type: element; text: link;
 spec: html; type: element; text: a;
+spec: html; type: element; text:script
 </pre>
 
 <!--
@@ -193,17 +194,17 @@ spec: html; type: element; text: a;
       with that <a>environment settings object</a> as their <a for=request>client</a>.
     </dd>
 
-    <dt><dfn>same-origin request</dfn></dt>
+    <dt><dfn>same-origin-referrer request</dfn></dt>
     <dd>
-      A {{Request}} <var>request</var> is a <strong>same-origin request</strong>
-      if <var>request</var>'s <a for=request>origin</a> and the <a for=url>origin</a> of
-      <var>request</var>'s <a for=request>current url</a> are <a lt="same origin">the same</a>.
+      A {{Request}} <var>request</var> is a <strong>same-origin-referrer request</strong> if the
+      <a for=url>origin</a> of <var>request</var>'s <a>referrerURL</a> and the <a for=url>origin</a>
+      of <var>request</var>'s <a for=request>current URL</a> are <a lt="same origin">the same</a>.
     </dd>
 
-    <dt><dfn>cross-origin request</dfn></dt>
+    <dt><dfn>cross-origin-referrer request</dfn></dt>
     <dd>
-      A {{Request}} is a <strong>cross-origin request</strong> if it is
-      <em>not</em> <a lt="same-origin request">same-origin</a>.
+      A {{Request}} is a <strong>cross-origin-referrer request</strong> if it is
+      <em>not</em> a <a>same-origin-referrer request</a>.
     </dd>
   </dl>
 </section>
@@ -247,9 +248,8 @@ spec: html; type: element; text: a;
   <h3 dfn export id="referrer-policy-no-referrer" oldids="referrer-policy-state-no-referrer">"<code>no-referrer</code>"</h3>
 
   The simplest policy is <a>"<code>no-referrer</code>"</a>, which specifies
-  that no referrer information is to be sent along with requests made from a
-  particular <a for=request lt=client>request client</a> to any <a for=/>origin</a>. The header will be
-  omitted entirely.
+  that no referrer information is to be sent along with requests to any
+  <a for=/>origin</a>. The header <a><code>Referer</code></a> will be omitted entirely.
 
   <div class="example">
     If a document at <code>https://example.com/page.html</code> sets a policy of
@@ -260,16 +260,18 @@ spec: html; type: element; text: a;
 
   <h3 dfn export id="referrer-policy-no-referrer-when-downgrade" oldids="referrer-policy-state-no-referrer-when-downgrade">"<code>no-referrer-when-downgrade</code>"</h3>
 
-  The <a>"<code>no-referrer-when-downgrade</code>"</a> policy sends a full URL, <a
-  href="#strip-url">stripped for use as a referrer</a>, along with requests from <a>environment
-  settings objects</a> whose <a for="environment settings object">HTTPS state</a> is "`modern`" to
-  a <a>potentially trustworthy URL</a>, and requests from <a for=request>clients</a> whose <a
-  for="environment settings object">HTTPS state</a> is not "`modern`" to any <a for=/>origin</a>.
+  The <a>"<code>no-referrer-when-downgrade</code>"</a> policy sends a request's full
+  <a>referrerURL</a> <a href="#strip-url">stripped for use as a referrer</a> for requests:
 
-  Requests from <a for=request>clients</a> whose <a for="environment settings object">HTTPS state</a> is "`modern`"
-  to non-<a>potentially trustworthy URL</a>s, on the other hand, will contain no
-  referrer information. A <code><a>Referer</a></code> HTTP header will not be
-  sent.
+  <ul>
+      <li>whose <a>referrerURL</a> and <a for=request>current URL</a> are both <a>potentially
+      trustworthy URLs</a>, or</li>
+      <li>whose <a>referrerURL</a> is a non-<a>potentially trustworthy URL</a>.</li>
+  </ul>
+
+  Requests whose <a>referrerURL</a> is a <a>potentially trustworthy URL</a> and whose
+  <a for=request>current URL</a> is a non-<a>potentially trustworthy URL</a> on the other hand, will
+  contain no referrer information. A <code><a>Referer</a></code> HTTP header will not be sent.
 
   <div class="example">
     If a document at <code>https://example.com/page.html</code> sets a policy of
@@ -288,12 +290,10 @@ spec: html; type: element; text: a;
 
   <h3 dfn export id="referrer-policy-same-origin">"<code>same-origin</code>"</h3>
 
-  The <a>"<code>same-origin</code>"</a> policy specifies that a
-  full URL, <a href="#strip-url">stripped for use as a referrer</a>, is sent as
-  referrer information when making <a>same-origin requests</a> from a particular
-  <a for=request>client</a>.
+  The <a>"<code>same-origin</code>"</a> policy specifies that a request's full <a>referrerURL</a> is
+  sent as referrer information when making <a>same-origin-referrer requests</a>.
 
-  <a>Cross-origin requests</a>, on the other hand, will contain no
+  <a>Cross-origin-referrer requests</a>, on the other hand, will contain no
   referrer information. A <code><a>Referer</a></code> HTTP header will not be
   sent.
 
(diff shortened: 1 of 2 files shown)


domfarolino:
The test https://github.com/web-platform-tests/wpt/pull/23552 has been merged, for the second bullet point in https://github.com/w3c/webappsec-referrer-policy/pull/135#issue-416916261.

index.src.html
```diff
@@ -193,17 +193,26 @@ <h2 id="terms">Key Concepts and Terminology</h2>
       with that <a>environment settings object</a> as their <a for=request>client</a>.
     </dd>
 
-    <dt><dfn>same-origin request</dfn></dt>
+    <dt><dfn>resolved referrer</dfn></dt>
     <dd>
-      A {{Request}} <var>request</var> is a <strong>same-origin request</strong>
-      if <var>request</var>'s <a for=request>origin</a> and the <a for=url>origin</a> of
-      <var>request</var>'s <a for=request>current url</a> are <a lt="same origin">the same</a>.
+      A {{Request}} <var>request</var>'s <strong>resolved referrer</strong> is a <a for=/>URL</a>
```
estark37, line 198: This term is a bit unintuitive to me; I would have never guessed what it meant before reading the definition. What do you think about "referrer source" or "referrer source URL"?

index.src.html
```diff
@@ -193,17 +193,26 @@ <h2 id="terms">Key Concepts and Terminology</h2>
...
+      that is the |referrerSource| variable in the [[#determine-requests-referrer]] algorithm. It is
```
estark37, line 199: How about something like: "The resolved referrer is the URL to which the referrer policy is applied. It may be set on the request directly or derived from request's client." ?

Also imo it would be more natural to have a note in the algorithm itself that says "referrerSource is the request's resolved referrer" than to reference the algorithm's variable here.

index.src.html
```diff
@@ -309,21 +319,27 @@ <h3 dfn export id="referrer-policy-same-origin">"<code>same-origin</code>"</h3>
     <a><code>Referer</code></a> header.
   </div>
 
+  <div class="example">
+    If a document at <code>https://example.com/page.html</code> sets a policy of
+    <a>"<code>same-origin</code>"</a>, and fetches a module script at
+    <code>https://script.example.com</code>, which then fetches a descendant script
+    at <code>https://example.com/descendant.js</code>, the request for the descendant
+    script would send no <a><code>Referer</code></a> header.
```
estark37, line 327: Could you perhaps expand this example to explain that the header will not be sent because the descendant script's current URL is https://example.com/descendant.js and its resolved referrer is https://script.example.com, hence the request is cross-origin-referrer?

index.src.html
```diff
@@ -374,22 +397,30 @@ <h3 dfn export id="referrer-policy-strict-origin">"<code>strict-origin</code>"</
     <code>http://example.com/</code>.
   </div>
 
+  <div class="example">
+    If a document at <code>http://example.com/page.html</code> sets a policy of
+    <a>"<code>strict-origin</code>"</a>, and fetches a module script at
+    <code><strong>https</strong>://script.example.com</code>, which then fetches a
+    descendant script at <code><strong>http</strong>://descendant.example.com</code>,
```
estark37, line 404: I think that almost all browsers would prevent this fetch from happening as a mixed script load (unless overridden by the user). It thus might be a slightly confusing example to include and perhaps not super useful (since it wouldn't come up often in real life). I'm open to being convinced otherwise though.

index.src.html
```diff
@@ -407,26 +438,32 @@ <h3 dfn export id="referrer-policy-origin-when-cross-origin" oldids="referrer-po
     <a>potentially trustworthy URL</a>s.
   </div>
 
+  <div class="example">
+    If a document a <code>https://example-1.com</code> sets a policy of
+    <a>"<code>origin-when-cross-origin</code>"</a>, and fetches a module script
+    at <code>https://example-2.com</code>, which then fetches a descendant script at
+    <code>https://example-1.com/descendant.js</code>, the request to the descendant
+    script would send a <a><code>Referer</code></a> header with a value of
+    <code>https://example-1.com/</code>.
```
estark37, line 447: Shouldn't this be example-2.com?

index.src.html
```diff
@@ -407,26 +438,32 @@ <h3 dfn export id="referrer-policy-origin-when-cross-origin" oldids="referrer-po
...
+  </div>
```
estark37, line 448: It might be worth having another example where the document is example-1.com, the module script is example-2.com, and the descendant script is also example-2.com

index.src.html
```diff
@@ -440,16 +477,24 @@ <h3 dfn export id="referrer-policy-strict-origin-when-cross-origin">"<code>stric
...
-    <code><strong>http://</strong>not.example.com/</code> would send no
+    <code><strong>http</strong>://not.example.com/</code> would send no
     <a><code>Referer</code></a> header.
   </div>
 
+  <div class="example">
+    If a document at <code>https://example.com/page.html</code> sets a policy of
+    <a>"<code>strict-origin-when-cross-origin</code>"</a>, and fetches a module script
+    at <code>https://script.example.com</code> which then fetches a descendant script at
+    <code><strong>http</strong>://descendant.example.com</code>, the request to the descendant
```
estark37, line 488: Same here about the mixed script load

index.src.html
```diff
@@ -865,22 +906,14 @@ <h3 id="determine-requests-referrer" dfn export>
         <dd>
           <ol>
             <li>
-              If <var>request</var> is a <a>same-origin request</a>, then
-              return <var>referrerURL</var>.
+              If the  <a for=url>origin</a> of |referrerURL| and the <a for=url>origin</a> of
+              <var>request</var>'s <a for=request>current url</a> are <a lt="same origin">the
+              same</a>, then return |referrerURL|.
```
estark37, line 911: optional nit: could add a note here saying that this determines whether the request is same-origin-referrer (to make the definition of same-origin-referrer completely concrete)

index.src.html
domfarolino: Thanks, I've made the following changes:
 - Removed this definition
 - Inside "determine request's referrer", when `referrerURL` is computed, I made that an actual `<dfn>` instead of just a variable, and associated it with the request, so it can be referenced outside of this algorithm (i.e., in the individual sections that explain what each policy does). I think this is much more clear and clean

If you prefer a different route, just let me know!

index.src.html
domfarolino: Thanks for the feedback, I think this is resolved given the comment on the below conversation.

index.src.html
domfarolino: That's what I thought initially, when I wrote tests for this a while ago, you can see in Chrome's console that if an secure module imports an insecure descendant, the descendant is not blocked. Visit http://wpt.live/html/semantics/scripting-1/the-script-element/module/referrer-strict-policies.sub.html and see the console in at least Chrome and Safari. The request doesn't seem to be blocked.

index.src.html
domfarolino: Done.

index.src.html
domfarolino: Heh, yep. Done.

index.src.html
domfarolino: (See my above reply about this)

index.src.html
domfarolino: Done.

index.src.html
domfarolino: Thanks, done.

domfarolino:
Thanks for the review @estark37, this is ready for another pass. Also, for some reason the build keeps failing (bikeshed fails to find the correct file to parse):

```
  File "/home/travis/build/w3c/webappsec-referrer-policy/bikeshed/bikeshed/Spec.py", line 274, in fixMissingOutputFilename
    elif self.inputSource.endswith(".src.html"):
AttributeError: 'FileInputSource' object has no attribute 'endswith'
```

But when I link locally, all is well, and there are no errors.

index.src.html
estark37: Ohh, I missed that the top-level document in this example is http://. I'd expect it to be blocked when the top-level document is https://, but not http://, so this makes sense.

estark37:
Thanks, LGTM! It looks like Travis wants to you to join the webappsec WG before submitting this (https://labs.w3.org/repo-manager/pr/id/w3c/webappsec-referrer-policy/135). Not sure what to do about the build failures... I just re-ran it to see if it magically fixed itself by chance.

domfarolino:
> I just re-ran it to see if it magically fixed itself by chance.

Thanks, yeah from some discussion in an internal chat, it seems that the latest Bikeshed that was causing the problem has been fixed, so it'll hopefully be good.

Regarding my membership...I'm working on it!

PR #137, opened 2020-06-08
Have no-referrer-when-downgrade specify stripping the URL as a referrer

davidvancleve:
Other policies that send "full" URLs specify explicitly that the URL should be stripped for use as a referrer. This change alters the description of the no-referrer-when-downgrade referrer policy to be consistent with the other policies in this regard.

***
<a href="https://pr-preview.s3.amazonaws.com/davidvancleve/webappsec-referrer-policy/pull/137.html" title="Last updated on Jun 11, 2020, 9:19 PM UTC (5bc9a80)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/137/fb230a1...davidvancleve:5bc9a80.html" title="Last updated on Jun 11, 2020, 9:19 PM UTC (5bc9a80)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -260,11 +260,11 @@ spec: html; type: element; text: a;
 
   <h3 dfn export id="referrer-policy-no-referrer-when-downgrade" oldids="referrer-policy-state-no-referrer-when-downgrade">"<code>no-referrer-when-downgrade</code>"</h3>
 
-  The <a>"<code>no-referrer-when-downgrade</code>"</a> policy sends a full URL, <a href="#strip-url">stripped for use as a referrer</a>, 
-  along with requests from <a>environment settings objects</a> whose <a for="environment settings object">HTTPS state</a>
-  is "`modern`" to a <a>potentially trustworthy URL</a>, and requests from
-  <a for=request>clients</a> whose <a for="environment settings object">HTTPS state</a> is not "`modern`"
-  to any <a for=/>origin</a>.
+  The <a>"<code>no-referrer-when-downgrade</code>"</a> policy sends a full URL, <a
+  href="#strip-url">stripped for use as a referrer</a>, along with requests from <a>environment
+  settings objects</a> whose <a for="environment settings object">HTTPS state</a> is "`modern`" to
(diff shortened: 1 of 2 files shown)


davidvancleve:
@mikewest PTAL

index.src.html
```diff
@@ -260,7 +260,7 @@ <h3 dfn export id="referrer-policy-no-referrer" oldids="referrer-policy-state-no
 
   <h3 dfn export id="referrer-policy-no-referrer-when-downgrade" oldids="referrer-policy-state-no-referrer-when-downgrade">"<code>no-referrer-when-downgrade</code>"</h3>
 
-  The <a>"<code>no-referrer-when-downgrade</code>"</a> policy sends a full URL
+  The <a>"<code>no-referrer-when-downgrade</code>"</a> policy sends a full URL, <a href="#strip-url">stripped for use as a referrer</a>,
```
domfarolino, line 263: Unfortunately this document doesn't really enforce a strict line-number wrapping, but I'd like to try and keep to 100 characters whenever changes are made - could you please wrap this?

davidvancleve: Done

davidvancleve:
PTAL

> Looks like we should be also including the generated `index.html` in this PR, could you generate that file and add it to this PR? (i.e., using #92 and #91 as reference)

Done

annevk:
This is a normative change that Firefox at least doesn't seem to implement. Did you file implementation bugs and write tests?

domfarolino:
I'm not sure how much of a behavior change this was - in the determine request's referrer algorithm, the [no-referrer-when-downgrade section](https://w3c.github.io/webappsec-referrer-policy/#ref-for-referrer-policy-no-referrer-when-downgrade%E2%91%A3) already returns the `referrerURL` variable, which was stripped for use as a referrer at the beginning of the algorithm, which never changed.

I think this PR just updates the description to match the behavior which already existed in the algorithm. I realize the descriptions are not marked as 'non-normative', so I guess it's a change in that sense, but the algorithm itself is the same. It would be good to have a test for this though.

PR #142, opened 2020-08-03
Make strict-origin-when-cross-origin the default referrer policy

krgovind:
This is a re-do of #125 by @mikewest as I will be taking over the relevant changes to the specs.

PRs against the fetch and html specs are coming.

***
<a href="https://pr-preview.s3.amazonaws.com/krgovind/webappsec-referrer-policy/pull/142.html" title="Last updated on Aug 3, 2020, 11:29 PM UTC (c4517e7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/142/d07eea4...krgovind:c4517e7.html" title="Last updated on Aug 3, 2020, 11:29 PM UTC (c4517e7)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -244,6 +244,8 @@ spec: html; type: element; text:script
   object</a> is used as a <a for=request lt=client>request client</a>. This policy may be tightened
   for specific requests via mechanisms like the <code><a>noreferrer</a></code>
   link type.
+  
+  The <dfn export>default referrer policy</dfn> is <a>"<code>strict-origin-when-cross-origin</code>"</a>.
 
   <h3 dfn export id="referrer-policy-no-referrer" oldids="referrer-policy-state-no-referrer">"<code>no-referrer</code>"</h3>
 
@@ -285,8 +287,6 @@ spec: html; type: element; text:script
     <code><strong>http</strong>://not.example.com/</code> would send no
     <a><code>Referer</code></a> header.
   </div>
-
-  This is a user agent's default behavior, if no policy is otherwise specified.
 
   <h3 dfn export id="referrer-policy-same-origin">"<code>same-origin</code>"</h3>
 
@@ -491,6 +491,9 @@ spec: html; type: element; text:script
     script would send no <a><code>Referer</code></a> header.
   </div>
 
+  This policy is the user agent's <a lt="default referrer policy">default</a>, and will be applied
+  if no policy is otherwise specified.
+
   <h3 dfn export id="referrer-policy-unsafe-url" oldids="referrer-policy-state-unsafe-url">"<code>unsafe-url</code>"</h3>
 
   The <a>"<code>unsafe-url</code>"</a> policy specifies that a request's full <a>referrerURL</a> is
@@ -513,9 +516,8 @@ spec: html; type: element; text:script
 
   The empty string "" corresponds to no <a for="/">referrer policy</a>, causing a
   fallback to a <a for="/">referrer policy</a> defined elsewhere, or in the case where
-  no such higher-level policy is available, defaulting to
-  <a>"<code>no-referrer-when-downgrade</code>"</a>. This defaulting happens in
-  the [[#determine-requests-referrer]] algorithm.
+  no such higher-level policy is available, falling back to the <a>default referrer
+  policy</a>. This happens in Fetch's main fetch algorithm, for example.
 
   <div class="example">
     Given a HTML <{a}> element without any declared <{a/referrerpolicy}>
@@ -1031,8 +1033,7 @@ spec: html; type: element; text:script
 
   Authors wanting to ensure that they do not leak any more information than
   the default policy should instead use the policy states
-  <a>"<code>same-origin</code>"</a>, <a>"<code>strict-origin</code>"</a>,
-  <a>"<code>strict-origin-when-cross-origin</code>"</a> or
+  <a>"<code>same-origin</code>"</a>, <a>"<code>strict-origin</code>"</a>, or
   <a>"<code>no-referrer</code>"</a>.
 
   <h3 id="downgrade">Downgrade to less strict policies</h3>
--- a/index.html
+++ b/index.html
@@ -992,57 +992,93 @@ Possible extra rowspan handling
 	.toc > li li          { font-weight: normal; }
 	.toc > li li li       { font-size:   95%;    }
 	.toc > li li li li    { font-size:   90%;    }
+	.toc > li li li li .secno { font-size: 85%; }
 	.toc > li li li li li { font-size:   85%;    }
+	.toc > li li li li li .secno { font-size: 100%; }
 
-	.toc > li             { margin: 1.5rem 0;    }
-	.toc > li li          { margin: 0.3rem 0;    }
-	.toc > li li li       { margin-left: 2rem;   }
+	/* @supports not (display:grid) { */
+		.toc > li             { margin: 1.5rem 0;    }
+		.toc > li li          { margin: 0.3rem 0;    }
+		.toc > li li li       { margin-left: 2rem;   }
 
-	/* Section numbers in a column of their own */
-	.toc .secno {
-		float: left;
-		width: 4rem;
-		white-space: nowrap;
-	}
-	.toc > li li li li .secno {
-		font-size: 85%;
-	}
-	.toc > li li li li li .secno {
-		font-size: 100%;
-	}
+		/* Section numbers in a column of their own */
+		.toc .secno {
+			float: left;
+			width: 4rem;
+			white-space: nowrap;
+		}
 
-	:not(li) > .toc              { margin-left:  5rem; }
-	.toc .secno                  { margin-left: -5rem; }
-	.toc > li li li .secno       { margin-left: -7rem; }
-	.toc > li li li li .secno    { margin-left: -9rem; }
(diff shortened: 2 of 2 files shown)


annevk:
Are we tracking limiting the ability for sites to opt in to looser policies somewhere?

krgovind:
> Are we tracking limiting the ability for sites to opt in to looser policies somewhere?

@annevk There is still some discussion that needs to happen on how to approach that correctly. I am hoping it can happen in PrivacyCG. Relevant thread: privacycg/proposals/issues/13

FYI, this was discussed on the [last WebAppSec call](https://github.com/w3c/webappsec/blob/master/meetings/2020/2020-07-21-minutes.md), and it was decided that the discussion should happen in PrivacyCG. There is also an open TAG review request at w3ctag/design-reviews/issues/538.

annevk:
@domfarolino just read your email on blink-dev. Note that since Safari ships a superset of this, I don't think we need to hold off here.

yoavweiss:
Safari's behavior seems to be [slightly different](https://groups.google.com/a/chromium.org/d/msg/blink-dev/lqFuqwZDDR8/L1eC6SGdAwAJ) that's the one described here.

I think it would be good to understand their reasoning, and see if we can converge on a single behavior.

yoavweiss:
^^ @johnwilander

Mr1262:
krgovind:defaultpolicychange

annevk:
To quote:

> Also, a clarification regarding Safari's current behavior: It appears that the policy "Third-party referrers are downgraded to origin..." is only applied to subresource requests and in the DOM; but not to the `Referer` request header on navigations and redirects.

That kind of discrepancy does seem problematic and in particular would not be a superset of this PR.

johnwilander:
I agree we should converge. Say we (WebKit) would downgrade on navigations and redirects. Would anyone else behave like us or are y’all still in the “just add a header and we’ll send full referrers again” camp? As I mentioned on a recent WebAppSec call, it seems these “change the default but with a new header/attribute you get the old behavior” changes we’re mostly creating busy work for developers. What is the goal?

annevk:
If if it's web-compatible I'd be in favor of "`strict-origin-when-cross-origin`" and aliasing "`unsafe-url`" and other less strict policies to it.

krgovind:
My 2 cents:

- Defaults matter. I think making the default referrer policy stricter is quite impactful, since a vast majority of developers don't specify a policy. Based on experience with changing the default across multiple browsers - experimentation on Chrome and Edge pre-stable channels, shipping in Firefox Private Browsing - it seems like a vast majority of sites won't need to make any changes to preserve user-visible functionality.
- Changing the default will help suss out use-cases where a full referrer is required, and allows us to be more deliberate in our approach. Note that there are other ways to pass along the referrer URL, other than simply specifying a permissive policy, such as via URL parameters.

johnwilander:
> My 2 cents:
> 
> * Defaults matter. I think making the default referrer policy stricter is quite impactful, since a vast majority of developers don't specify a policy. Based on experience with changing the default across multiple browsers - experimentation on Chrome and Edge pre-stable channels, shipping in Firefox Private Browsing - it seems like a vast majority of sites won't need to make any changes to preserve user-visible functionality.

I think there's an important distinction between settings under the user's control and settings under the site's control when we talk about "defaults matter." In the case of bad TLS, the default is to block and the user can override to relax. The site can only override to get stricter treatment (HSTS). That's Safari's "default matter" view in the case of referrers – the site can get a stricter policy if it wants but it cannot relax the policy.

Take SameSite=lax cookies by default for instance. The result seems to have been a lot of site owner work and angst and then back to full cross-site tracking because all trackers have set SameSite=none. WebKit keeps getting bugs filed where developers say "I've set SameSite=none and still don't get third-party cookies." They think they did the busy work needed and I don't think they'll be happy when browsers other than Safari+Brave+Tor come around to blocking all third-party cookies, presumably just two years after the SameSite=lax change.

> * Changing the default will help suss out use-cases where a full referrer is required, and allows us to be more deliberate in our approach. Note that there are other ways to pass along the referrer URL, other than simply specifying a permissive policy, such as via URL parameters.

I think we've received 2-3 bug reports in total about downgrading referrers and they were all fixed by the site owners IIRC. Mozilla who did the original research saw very little breakage too. Maybe it would be better to treat those few cases as quirks, since they might not be able/willing to reply an explicit referrer policy?

krgovind:
> I think there's an important distinction between settings under the user's control and settings under the site's control when we talk about "defaults matter." In the case of bad TLS, the default is to block and the user can override to relax. The site can only override to get stricter treatment (HSTS). That's Safari's "default matter" view in the case of referrers – the site can get a stricter policy if it wants but it cannot relax the policy.

Sorry, I'm not sure if I made it clear in my earlier response: Chrome is interested in considering capping the referrer as a long-term measure, but I'm asking that we think about the implications to known use-cases and how we expect developers to respond. Some usecases needing the full referrer URL such as contextual advertising and brand-safety in advertising were discussed on the [7/21 WebAppSec call](https://github.com/w3c/webappsec/blob/master/meetings/2020/2020-07-21-minutes.md). 

The presumable response to referrer capping is for developers working w/ these use cases to pass on the referrer as a URL parameter to the ad request, which defeats the purpose of capping. I don't think referrer capping by itself is a comprehensive measure.

> Take SameSite=lax cookies by default for instance. The result seems to have been a lot of site owner work and angst and then back to full cross-site tracking because all trackers have set SameSite=none. WebKit keeps getting bugs filed where developers say "I've set SameSite=none and still don't get third-party cookies." They think they did the busy work needed and I don't think they'll be happy when browsers other than Safari+Brave+Tor come around to blocking all third-party cookies, presumably just two years after the SameSite=lax change.

What you describe around developer confusion regarding SameSite and filing of WebKit bugs is I think due to an unfortunate coincidence of timing between the Chrome SameSite rollout and Safari ITP blocking 3p cookies by default.

I think blocking third-party cookies outright may actually be a counter-example here and is a cautionary tale to approach referrer capping more thoughtfully. In the absence of good solutions for their use-cases, developers have recommended that users [disable ITP entirely](https://cs.myjcom.jp/knowledgeDetail?an=002309185), or use a [different browser](https://www.reddit.com/r/PS4/comments/fxt9sk/safari_is_not_supported_sony_are_you_okay/).

> I think we've received 2-3 bug reports in total about downgrading referrers and they were all fixed by the site owners IIRC. Mozilla who did the original research saw very little breakage too. Maybe it would be better to treat those few cases as quirks, since they might not be able/willing to deploy an explicit referrer policy?

Could you clarify whether in those cases, the fix was for the site to make do with the origin? Were there any use-cases that required the full URL? We know of at least one case with a payment vendor where the referrer was being used as a fraud protection measure, but we have advised them to use the origin instead.

domfarolino (approved):
The changes here look good ~but I think we'll need a PR to fetch to change https://fetch.spec.whatwg.org/#ref-for-concept-request-referrer-policy%E2%91%A3, do you agree? Also we should do a pass over the HTML Standard to see if there is any mention of historical defaults etc~

I like that we consolidate the "default" in a single place, instead of having multiple specs reference what they believe is the default policy.

domfarolino:
Chrome is currently working on the WPT changes for this, so I'll hold-off on merging at least until then. I think it's OK to land this in general since from my understanding, the change adheres to the changes Firefox + Chrome aims to make. It also brings the spec _closer_ to the non-standard behavior that Safari shipped, so discussion aimed at getting even more conformance between those impls and the spec can happen asynchronously.

Tracking bugs below:

 - [Corresponding Fetch Standard PR](https://github.com/whatwg/fetch/pull/1066)
 - [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1014207) ([WPTs](https://bugs.chromium.org/p/chromium/issues/detail?id=1131688))
 - [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1589074)

krgovind:
@domfarolino Can we merge this PR now that the WPTs have landed?

domfarolino:
I think so, yes. Let me see if I can get an approval out of @annevk first if that's OK, just for extra assurance.

domfarolino:
Over IRC he's indicated he's alright landing this, so I'll go ahead and merge. Thanks for all of the work and discussion here everyone!

OBITOONDEADO1MZ (approved):

DuschSchmidt27 (approved):

index.src.html
```diff
@@ -286,8 +288,6 @@ <h3 dfn export id="referrer-policy-no-referrer-when-downgrade" oldids="referrer-
     <a><code>Referer</code></a> header.
   </div>
 
-  This is a user agent's default behavior, if no policy is otherwise specified.
```
DuschSchmidt27, line 289: "note": "Don't delete this file! It's used internally to help with page regeneration."

DuschSchmidt27, line 289: "note": "Don't delete this file! It's used internally to help with page regeneration."

DuschSchmidt27 (changes requested):

index.src.html
DuschSchmidt27, line 289: "note": "Don't delete this file! It's used internally to help with page regeneration."

DuschSchmidt27 (changes requested):

PR #147, opened 2021-01-19
Use GitHub workflows to publish the spec

domfarolino:
This PR ports over the `build.yml` and `Makefile` from some other specs that use a GitHub-actions based auto-publish workflow.

The benefit of this change is that contributors no longer have to include changes to the output spec file in their PRs, which is a messy business. Instead the spec will be auto-built and published to the gh-pages branch.


Makefile
```diff
@@ -1,11 +1,28 @@
...
+
+clean:
+	rm index.html
+
+index.html: index.bs
+	@ (HTTP_STATUS=$$(curl https://api.csswg.org/bikeshed/ \
+	                       --output index.html \
+	                       --write-out "%{http_code}" \
+	                       --header "Accept: text/plain, text/html" \
+												 -F die-on=warning \
```
domenic, line 22: Indents look off here

domenic, line 22: This spec probably hasn't been building with "die-on=warning" so far so you might want to omit this (both here and on line 6) until it's building cleanly without warnings.

Actually it looks like the spec is building with `-f` (`--force`) which causes certain errors to be ignored. So you might need to do that too. (It'd be `-F force=1` for this line.)

domenic (changes requested):
I would suggest not renaming the `index.src.html` file since that will lose history, at least on GitHub. (Not on command-line `git` though.) Instead just update the makefile by replacing `index.bs` with `index.src.html` throughout.

Makefile
domfarolino: That's really weird. I copied this directly from https://github.com/jeremyroman/alternate-loading-modes. Fixed.

domfarolino: As discussed on chat, die-on=warning works here which is great, so we'll stick with this.

domfarolino:
> I would suggest not renaming the index.src.html file since that will lose history, at least on GitHub. (Not on command-line git though.) Instead just update the makefile by replacing index.bs with index.src.html throughout.

Yeah that's reasonable, I just happen to like the bs file extension a lot more. Done.

domfarolino:
This PR also includes deleting `.travis.yml` to stop travis from attempting to build the spec, since we're handling that here now.

domenic (approved):
LGTM. You may also want to try deleting biblio.json and seeing if `make local` still works, since make remote` and `make ci` don't use it.

PR #148, opened 2021-01-21
Clarify the rare inheritance of referrer policy

domfarolino:
This PR replaces the "Nested browsing context" section with a more general "Referrer Policy Inheritance" section that outlines what I see as the only two places where a referrer policy is ever effectively inherited:
 - For requests coming from an iframe srcdoc document, and...
 - During `javascript:` navigations

This section was previously wrong because it mentions that we only inherit the referrer policy for documents not created by a response, like srcdoc documents, however srcdoc documents are actually created from a (locally-created) response, so the previous text wasn't quite right. It also leaves out `javascript:` navigations which use locally-created responses as well when the JS evaluates to a string.

This section also used to mention "blob:" URLs, however from looking around in HTML as well as https://fetch.spec.whatwg.org/#concept-scheme-fetch I do not see any special referrer policy treatment in the "blob:" URL case ... maybe @annevk would know.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/pull/148.html" title="Last updated on Jan 22, 2021, 6:48 PM UTC (a03517d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/148/828435e...a03517d.html" title="Last updated on Jan 22, 2021, 6:48 PM UTC (a03517d)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -648,7 +648,7 @@ spec: html; type: element; text:script
     <em>This section is not normative.</em>
 
     {{Document}}s don't typically inherit their <a for="Document">referrer policy</a> from other
-    {{Document}}s. However, there are two exceptions:
+    {{Document}}s. However, there are currently three cases where this happens:
     - Requests that do not provide their own <a for="request">referrer policy</a> and are initiated
       from <a>an <code>iframe</code> <code>srcdoc</code> document</a> that does not set its own
       <a for="Document">referrer policy</a> (via a <{meta}> tag) will use the
@@ -671,6 +671,9 @@ spec: html; type: element; text:script
           the first point above in this section.
         - For all other <a>browsing contexts</a>, the new {{Document}}'s <a for="Document">referrer
           policy</a> is inherited from the {{Document}} preceding the "<code>javascript:</code>" navigation.
+    - Upon <a spec=html lt="create a new browsing context">creation</a> of a new <a>browsing context</a>,
+      the initial <code><a spec=html>about:blank</a></code> {{Document}} will inherit its
+      <a for="Document">referrer policy</a> from its creator {{Document}} if the creator is non-null.
 
     Note: When a Policy Container is fully defined, the inheritance of <a for="/">referrer policies</a> or
     lack thereof will be more consistent and clear, as opposed to relying on subtle edge cases described above.
--- a/index.src.html
+++ b/index.src.html
@@ -643,7 +643,7 @@ spec: html; type: element; text:script
 
   <section class="informative">
     <h3 id="referrer-policy-inheritance"
-    oldids="referrer-policy-delivery-nested, referrer-policy-delivery-implicit"><a for="/">Referrer policy</a> Inheritance</h3>
+    oldids="referrer-policy-delivery-nested, referrer-policy-delivery-implicit"><a for="/">Referrer Policy</a> Inheritance</h3>
 
     <em>This section is not normative.</em>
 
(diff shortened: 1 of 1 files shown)


domfarolino:
@domenic If you could take a look too that'd be great. Looks like I can't add you as a reviewer though, so let me figure that out.

domenic (commented):
Content LGTM with editorial nits. However I worry about the blob case. I think there might be open issues on the File API, or HTML, about getting that specified? Maybe just the "policy container" issue on HTML?

It might be best not to merge this until figuring out the story there, either to add a pointer to some existing spec we missed, or a pointer to an open issue.

annevk:
What about initial `about:blank`, `about:blank`, `blob:` URLs, `data:` URLs, and `document.open()` (maybe no longer applicable)? This seems better tackled as part of https://github.com/antosart/policy-container-explained indeed.

domfarolino:
Yeah from talking to @antosart it seems like Policy Container work will result in copying the navigation initiator's Policy Container upon navigation to a local scheme. I'm happy to point to the Policy Container issue on HTML for now with a note saying that Policy Container will handle this in a more cohesive and consistent way.

domenic (approved):
LGTM since it reflects the state of all well-specified things. I did find

- https://wpt.fyi/results/referrer-policy/generic/inheritance/iframe-inheritance-about-blank.html?label=experimental&label=master&aligned
- https://wpt.fyi/results/referrer-policy/generic/inheritance/iframe-inheritance-data.html?label=master&label=experimental&aligned

although I didn't dig deep enough to understand what those were testing.

domfarolino:
Ah, it looks like I wrote those tests last year and you reviewed the, but I totally forgot about them here. Yes, the other well-specified example is when a new browsing context is created with a non-null creator, the initial `about:blank` Document will:
 - Inherit the creator Document's referrer policy
 - Set its own `document.referrer` to the unredacted URL of the creator Document

I'll mention this in the spec PR and land it.

annevk:
There's also the case of navigating something to `about:blank` (i.e., non-initial `about:blank`).

domfarolino:
> There's also the case of navigating something to about:blank (i.e., non-initial about:blank).

Right, but I thought navigating to `about:blank` would not inherit the referrer policy in this case... hmm but let me check

annevk:
I would expect it to inherit from whoever did the navigating. That's what we do for other policies.

domfarolino:
Hmm, I don't think that's very common referrer policy at all though. Navigating to `about:blank` or any other URL will ultimately [create and initialize](https://html.spec.whatwg.org/multipage/browsing-the-web.html#initialise-the-document-object) a new Document, and its referrer policy is set in (currently) step 9 when we parse the `Referrer-Policy` header of the navigation response. So it is not inherited from the initiator, unless the response has an artificially-loaded `Referrer-Policy` header, like responses created from `javascript:` navigations (see this PR).

Does that make sense?

domfarolino:
I think that's exactly what Policy Container sets out to solve

annevk:
It makes sense that things are currently broken. I guess I wonder if the same is true for implementations. 😊

domfarolino:
Right. I ended up filing some bugs at least against Chromium for this last year I believe, relating to the tests Domenic pointed out above. It sounds like much of this will get addressed with Policy Container which will attempt to tidy up the area a lot.

I do think that for referrer policy, we should _only_ ever inherit the referrer policy from a navigation initiator for local schemes. Otherwise, if A.com sets referrer policy `unsafe-url` and it navigates to B.com which is delivered with a header `no-referrer` or is not delivered with a header, I don't think B.com should just get `unsafe-url` as its policy. That seems wrong.

Issue #155, opened 2021-11-09
Omit referrers on cross-origin requests from an RFC7686 address


fmarier:
The special-use `.onion` domain name, defined in [RFC7686](https://datatracker.ietf.org/doc/rfc7686/), receives special treatment in the Tor browser and in Firefox: https://searchfox.org/mozilla-central/rev/f8576fec48d866c5f988baaf1fa8d2f8cce2a82f/dom/security/ReferrerInfo.cpp#334-339

It seems like this behavior should be standardized since any browser could be setup to proxy traffic over the Tor SOCKS5 proxy.

mikewest:
Is the `StaticPrefs::network_http_referer_hideOnionSource()` preffed on by default?

It looks like Mozilla's implementation doesn't allow a page's assertion of a laxer policy to override the `.onion` === `none` behavior. I wonder whether there are cases where the site wouldn't want that. Referrer-based ACLs, etc?

annevk:
That's a good question, Firefox does not ship this by default: https://searchfox.org/mozilla-central/source/modules/libpref/init/StaticPrefList.yaml#9297-9302.

fmarier:
Good catch, I hadn't noticed that this wasn't ON by default in Firefox, just in the Tor Browser. I just tested this with the [test page I've been using](http://ixrdj3iwwhkuau5tby5jh3a536a2rdhpbdbu6ldhng43r47kim7a3lid.onion/referrer/onion.html) to replicate the Tor Browser behavior in Brave and Firefox doesn't touch these headers when set to use a local Tor daemon as a SOCKS5 proxy. I updated the two-implementer note on the [Fetch PR](https://github.com/whatwg/fetch/pull/1351) accordingly.

fmarier:
> I wonder whether there are cases where the site wouldn't want that. Referrer-based ACLs, etc?

It seems unlikely to me that a site would rely on receiving a `.onion` referrer for any reason because the majority of users coming from a `.onion` site will be using the Tor Browser, which doesn't send a referrer in this case. Unless it's to _exclude_ users coming from a `.onion` site, but I don't know what the point of that would be.

Issue #154, opened 2021-11-09
Bikeshed (remote) returns an error on main branch


fmarier:
```
$ make remote
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 50415    0   791  100 49624    410  25725  0:00:01  0:00:01 --:--:-- 26121

Error running preprocessor, returned code: 2.
LINK ERROR: Multiple possible 'policy containers' dfn refs.
Arbitrarily chose https://fetch.spec.whatwg.org/#concept-request-policy-container
To auto-select one of the following refs, insert one of these lines into a &lt;pre class=link-defaults> block:
spec:fetch; type:dfn; text:policy container
spec:html; type:dfn; for:Document; text:policy container
spec:html; type:dfn; for:/; text:policy container
spec:html; type:dfn; for:environment settings object; text:policy container
spec:html; type:dfn; for:WorkerGlobalScope; text:policy container
spec:service-workers; type:dfn; text:policy container
&lt;a data-link-type="dfn" data-lt="policy containers">policy containers&lt;/a>
 ✘  Did not generate, due to fatal errors
make: *** [Makefile:18: index.html] Error 22
```

annevk:
@antosart looks like there is some breakage due to policy container changes here.

PR #156, opened 2021-11-09
Omit referrers on cross-origin requests from an .onion address

fmarier:
This is to address #155.

It's currently a draft PR because I haven't been able to test generating the spec HTML due to the current version failing to link with Bikeshed (#154).


annevk:
https://github.com/whatwg/fetch/pull/1351 proposes similar language. Do we expect many other places to need this? If we find a third it probably warrants abstraction.

fmarier:
> Do we expect many other places to need this? If we find a third it probably warrants abstraction.

The only other special treatment of `.onion` I could see [in Firefox](https://searchfox.org/mozilla-central/search?q=.onion&case=true&path=) has to do with treating `.onion` services over HTTP as a secure context, which we are also [planning to do in Brave](https://github.com/brave/brave-browser/issues/1135).

index.src.html
```diff
@@ -850,6 +850,13 @@ <h3 id="determine-requests-referrer" dfn export>
       Let request's <dfn>referrerURL</dfn> be the result of <a href="#strip-url">stripping
       <var>referrerSource</var> for use as a referrer.</a>
     </li>
+    <li>
+      If the <a for=url>origin</a> of <a>referrerURL</a> has a <a for=origin>host</a> that
+      ends with <code>.onion</code> or <code>.onion.</code>, the special-use domain name defined
+      in [[!RFC7686]], and that <a for=url>origin</a> is not <a lt="same origin">the same</a> as
+      the <a for=url>origin</a> of <var>request</var>'s <a for=request>current URL</a>, then
```
domfarolino, line 857: Could we just use the https://w3c.github.io/webappsec-referrer-policy/#cross-origin-referrer-request dfn here?

fmarier: Indeed. That reduces the unnecessary verbosity.

domfarolino:
This has been sitting for a while now with no activity (partially because I've failed to review it, but that's mostly because I haven't seen much activity in terms of consensus on the other related issues). I've tried to round up all of the related web platform `.onion` issues and PRs:

 - https://github.com/whatwg/fetch/issues/1350
 - https://github.com/w3c/webappsec-referrer-policy/issues/155
 - https://github.com/whatwg/fetch/pull/1351
 - (This PR)
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1742784

It seems like the Fetch issue is stalled on feedback that hasn't come in, and perhaps also stalled on general consensus for this from other browsers. Would you mind if I closed this or marked it as a draft until some of the consensus around handling `.onion` in web specs gets sorted out, and we can go from there? Let me know if you have plans to still work on this.

fmarier:
You can also add https://github.com/brave/brave-browser/issues/18071 if you want (that's the Brave issue where we implemented this in order to match the Tor Browser behavior).

Marking it as draft makes sense. I do intend to pick this up again since I think it would be worthwhile to agree on what the correct behavior should be and/or standardize what the two browser with built-in support for Tor already ship.

PR #162, opened 2022-05-24
Fix 'Strip url for use as referrer'

### LINKED ISSUE #153: "Strip url for use as a referrer" sets path to null, which is a spec type error
c.f. https://w3c.github.io/webappsec-referrer-policy/#strip-url and https://url.spec.whatwg.org/#concept-url-path

Notably, Referrer Policy says:

> Set url’s path to null.

But URL says:

> A URL’s path is a list of zero or more ASCII strings, usually identifying a location in hierarchical form. It is initially empty.

(i.e., doesn't permit null as a value)

Looking at http://wpt.live/referrer-policy/4K+1/gen/top.meta/same-origin/xhr.http.html, it appears that Chrome uses `[""]` (resulting in `http://wpt.live/`) and Safari uses `[]` (resulting in `http://wpt.live`). Firefox seems to just use no-referrer in this case.

domfarolino:
This closes #153.

 - Set url's "password" to an empty string instead of null, since null is not an allowable type per the URL Standard
 - Same with "path"
 - Adds `for=url` on several `<a>` in this algorithm, since "fragment" was linking to HTML's infrastructure page

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/pull/162.html" title="Last updated on May 24, 2022, 12:26 AM UTC (3321fcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/162/a02b69a...3321fcd.html" title="Last updated on May 24, 2022, 12:26 AM UTC (3321fcd)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -987,7 +987,7 @@ spec: html; type: element; text:script
 
       <ol>
         <li>
-          Set <var>url</var>'s <a for=url>path</a> to « the empty string ».
+          Set <var>url</var>'s <a for=url>path</a> to « ».
         </li>
         <li>
           Set <var>url</var>'s <a for=url>query</a> to <code>null</code>.
--- a/index.src.html
+++ b/index.src.html
@@ -56,7 +56,6 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: browsing context
     urlPrefix: infrastructure.html
       text: ascii case-insensitive match; url: ascii-case-insensitive
-      text: fragment; url: concept-url-fragment
       text: document base url
       text: plugin
       text: reflect
--- a/index.src.html
+++ b/index.src.html
@@ -974,13 +974,13 @@ spec: html; type: element; text:script
       return <code>no referrer</code>.
     </li>
     <li>
-      Set <var>url</var>'s <a>username</a> to the empty string.
+      Set <var>url</var>'s <a for=url>username</a> to the empty string.
     </li>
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -974,21 +973,21 @@ <h3 id="strip-url">
...
+      Set <var>url</var>'s <a for=url>fragment</a> to <code>null</code>.
     </li>
     <li>
       If the <code><a>origin-only flag</a></code> is <code>true</code>,
       then:
 
       <ol>
         <li>
-          Set <var>url</var>'s <a for=url>path</a> to <code>null</code>.
+          Set <var>url</var>'s <a for=url>path</a> to « the empty string ».
```
domenic, line 990: Should this be the empty list?

domfarolino: Hmm, it looks like it. I was on autopilot looking at https://github.com/w3c/webappsec-referrer-policy/issues/153#issuecomment-864359547.

domfarolino: Done. I've made this `« »`, but let me know if you think it should be "an empty list".

domenic (approved):

annevk:
Unfortunately this is wrong as I noted in #153. The empty list you end up with no `/`, but there should be a `/`. It really should be

> « the empty string »

as I said.

See also https://url.spec.whatwg.org/#url-path-serializer.

PR #169, opened 2023-11-24
Use cross-spec section autolinks instead of absolute URLs

tidoust:
This is purely editorial. The spec had a couple of places where it referenced a section of other specs (RFC9110 and Fetch) using absolute URLs. This update rather uses the recommended cross-spec section autolink: https://speced.github.io/bikeshed/#autolink-biblio

This will allow Bikeshed to detect and alert you should the fragment identifiers change in the targeted specifications, as well as to follow the specification if its URL changes (the cross-references database typically now uses the `httpwg.org/specs/` URL for RFC9110 as it is slightly more readable).


tidoust:
Note I was a bit too prompt to create the PR. The new data has not yet reached the [online Bikeshed service](https://api.csswg.org/bikeshed/) that the build code uses to generate the spec. That should happen soon...

index.src.html
```diff
@@ -569,8 +569,8 @@ <h3 id="referrer-policy-header">Delivery via Referrer-Policy header</h3>
 
   The syntax for the name and value of the header are described by the following
   ABNF grammar. ABNF is defined in [[!RFC5234]], and the <code>#rule</code> ABNF
-  extension used below is defined in <a
-  href="https://www.rfc-editor.org/rfc/rfc9110#section-5.6.1">Section 5.6.1</a> of [[!RFC9110]].
+  extension used below is defined in [[RFC9110#abnf.extension|Section 5.6.1]] of
```
domfarolino, line 572: This link does not show up in the preview https://pr-preview.s3.amazonaws.com/tidoust/webappsec-referrer-policy/pull/169.html#referrer-policy-header

domfarolino: This seems to be the cause of the build error too.

tidoust: Yes, see my previous comment: https://github.com/w3c/webappsec-referrer-policy/pull/169#issuecomment-1825858988

The new data has now reached the online Bikeshed service. I re-ran the CI build, which now passes. I don't think I can run PR preview again though.

domfarolino:
I think this should be good to go now, but you probably need to rebase it on top of the new main head for the build to succeed, since when I re-ran it, there was a build failure that I have just now fixed. Could you do that?

Issue #168, opened 2023-06-27
Add Referrer-Policy to HTTP Field Name Registry


stiiin:
`Referrer-Policy` is not listed in IANA's [Hypertext Transfer Protocol (HTTP) Field Name Registry](https://www.iana.org/assignments/http-fields/http-fields.xhtml). I'm just making sure that there's an issue tracking this.

stiiin:
`Referrer-Policy` is now listed in the [HTTP Field Name Registry](https://www.iana.org/assignments/http-fields/http-fields.xhtml) :)

PR #171, opened 2024-08-23
Added IANA considerations section for HTTP header

stiiin:
This change adds an IANA Considerations section, with the intent to register the Referrer-Policy HTTP header. This should move #168 along.

-@stiiin

***
<a href="https://pr-preview.s3.amazonaws.com/stiiin/webappsec-referrer-policy/pull/171.html" title="Last updated on Aug 24, 2024, 5:06 PM UTC (0d40542)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/171/c7db4d9...stiiin:0d40542.html" title="Last updated on Aug 24, 2024, 5:06 PM UTC (0d40542)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1078,6 +1078,29 @@ spec: html; type: interface; text:Window
 
 </section>
 
+<section>
+  <h2 id="iana">IANA Considerations</h2>
+
+  The HTTP Field Name Registry is to be updated with the following
+  registration: [[!RFC9110]]
+
+  <dl>
+    <dt>Field name</dt>
+    <dd>Referrer-Policy</dd>
+
+    <dt>Status</dt>
+    <dd>permanent</dd>
+
+    <dt>Specification document(s)</dt>
+    <dd>This specification (See [[#referrer-policy-header]])</dd>
+
+    <dt>Comments</dt>
+    <dd>
+      The header name does not share the HTTP Referer header's misspelling.
+    </dd>
+  </dl>
+</section>
+
 <!--
    ███     ██████  ██    ██ ██    ██  ███████  ██      ██ ██       ████████ ████████   ██████   ████████ ██     ██ ████████ ██    ██ ████████  ██████
   ██ ██   ██    ██ ██   ██  ███   ██ ██     ██ ██  ██  ██ ██       ██       ██     ██ ██    ██  ██       ███   ███ ██       ███   ██    ██    ██    ██


stiiin:
Note: I think I've fumbled the requested status of the registration a bit.

It makes sense to a layperson like me to request a `permanent` status from a specification. But because I'm just a random bystander, I figured [my registration request](https://github.com/protocol-registries/http-fields/issues/50) should be `provisional`.

If someone with more experience on this administrative work could tell me which status would be correct, I'll happily update either this PR or the registration request.

domfarolino:
I'm not really familiar with the whole IANA registration process and overlap with web platform specifications. Is it common for web specifications to have a considerations section that defines new headers that have not yet been added formally?

index.src.html
```diff
@@ -1078,6 +1078,31 @@ <h3 id="unknown-policy-values">Unknown Policy Values</h3>
 
 </section>
 
+<section>
+  <h2 id="iana">IANA Considerations</h2>
+
+  <h3 id="iana-http-headers">HTTP Headers</h3>
+
+  The HTTP Field Name Registry should be updated with the following
```
domfarolino, line 1086: Maybe "is to be updated" instead of "should", given the meaning of "should" per https://infra.spec.whatwg.org/#conformance.

annevk:
Some people like IANA sections in specifications, but you can also just register directly without doing that. That's what Fetch does for instance.

domfarolino:
Ah, yeah I personally think that is best then. Given the request, I'm thinking we can close this. Is that OK with you @stiiin ?

stiiin:
I'm OK with any action that helps to get the registry updated! :)

The reason I made this PR was because @mikewest [suggested this way forward](https://lists.w3.org/Archives/Public/public-webappsec/2024Aug/0004.html), in response to @simoneonofri [contacting the working group](https://github.com/protocol-registries/http-fields/issues/50#issuecomment-2296175962). Since Mike mentioned two former editors, as well as mentioning that he's on vacation himself, I figured I might as well submit a patch ahead of time.

index.src.html
```diff
@@ -1078,6 +1078,31 @@ <h3 id="unknown-policy-values">Unknown Policy Values</h3>
 
 </section>
 
+<section>
+  <h2 id="iana">IANA Considerations</h2>
+
+  <h3 id="iana-http-headers">HTTP Headers</h3>
```
mikewest, line 1084: Nit: I think this header is unnecessary, given that there's only one thing we're asking IANA to do. CSP had several, AFAIR, so multiple headers were helpful.

mikewest (approved):
I think adding this is pretty reasonable, but I don't have strong feelings about it if it's not actually required. My (very vague!) understanding was that each document's IANA instructions were the formal justification for the registration, but if that's unnecessary I'm happy to avoid it here (and, I suppose, remove the relevant sections from other webappsec specs).

I'd defer to @mnot and @simoneonofri for IETF and W3C opinions, respectively. If they deem this unnecessary, let's drop it. If not, I'm happy with the patch modulo the nit I left as a comment.

mnot:
From the registry expert standpoint: I don't need it in the document (although that's nice just for posterity). What's important is that the WG is aware of and OK with the request.

simoneonofri:
[it make sense for me](https://github.com/protocol-registries/http-fields/issues/50#issuecomment-2296175962) and WG is [aware of the registration](https://lists.w3.org/Archives/Public/public-webappsec/2024Aug/0004.html)

domfarolino:
Likewise. If @mikewest doesn't have strong opinions, I vote we proceed with the request as usual, and close this PR out since the changes here are not needed for the request to proceed.

stiiin:
Closing as suggested.

PR #177, opened 2025-12-25
Use API base URL when determining referrer

TimvdLippe:
Based on an investigation as to why Servo
doesn't use the correct URL per the spec [1]
I thought that HTML was incorrectly setting
the creation URL [2].

However, after writing a WPT test [3] it turns
out that the creation URL is correctly set in
HTML, but incorrectly used in Referrer Policy.
Browsers don't use the creation URL to
determine the referrer policy, per another
WPT test [4].

Instead of the creation URL, browsers use the
url of the worker, which is the same as the
API base URL [5].

[1]: https://github.com/servo/servo/pull/41458
[2]: https://github.com/whatwg/html/pull/12038
[3]: https://github.com/web-platform-tests/wpt/pull/56921
[4]: https://wpt.fyi/results/fetch/api/basic/request-referrer-redirected-worker.html?label=master&label=experimental&aligned
[5]: https://html.spec.whatwg.org/multipage/workers.html#script-settings-for-workers:api-base-url

***
<a href="https://pr-preview.s3.amazonaws.com/TimvdLippe/webappsec-referrer-policy/pull/177.html" title="Last updated on Dec 25, 2025, 10:22 AM UTC (b60f97a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-referrer-policy/177/2adab47...TimvdLippe:b60f97a.html" title="Last updated on Dec 25, 2025, 10:22 AM UTC (b60f97a)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -828,8 +828,8 @@ spec: html; type: interface; text: Window
               </ol>
             </li>
 
-            <li>Otherwise, let |referrerSource| be |environment|'s <a for=environment>creation
-            URL</a>.</li>
+            <li>Otherwise, let |referrerSource| be |environment|'s
+              <a for="environment settings object">API base URL</a>.</li>
           </ol>
         </dd>
 


domfarolino:
> However, after writing a WPT test https://github.com/web-platform-tests/wpt/pull/56921 it turns out that the creation URL is correctly set in HTML

Can you clarify what you mean by it being "correctly" set in HTML? In https://github.com/whatwg/html/pull/12038 you mentioned that a Worker's creation URL lags behind redirects, since it's never updated after the Fetch; you indicated that this was an error so you strove to update it. But now you're saying that your HTML patch is wrong, and that the creation URL is actually correct. But to me, it still seems like the Worker creation URL can desync from the final resource's URL, right?



> Instead of the creation URL, browsers use the
url of the worker, which is the same as the
API base URL [5](https://html.spec.whatwg.org/multipage/workers.html#script-settings-for-workers:api-base-url).

We cannot just use the API base URL as a blanket policy here. For Window environments, [the API base URL is influenced by `<base>` elements](https://html.spec.whatwg.org/#script-settings-for-window-objects:api-base-url), and as far as I can tell, the referrer generated for relative network requests in a Document never considers `<base>` elements. Consider the following test:

```js
const base = document.createElement('base');
base.href = 'https://example.com/test/';
document.head.append(base);
new Image().src = 'relative'; // Request is to https://example.com/test/relative, with referrer https://example.com
```

No browser sends a request to `/test/relative` with the referrer `/test/`, but the spec would after this PR. I'm wondering if we should just go the "fix the creation URL route" after all... thoughts?

TimvdLippe:
Window environments are handled separately in step 3."client".2 of https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer Therefore, we only use the creation URL for non-window global scopes (which are the workers).

> But to me, it still seems like the Worker creation URL can desync from the final resource's URL, right?

That's correct. The `creation URL` is the initial URL, the `url` is the final URL considering redirects. Other specifications such as the reporting API rely on the creation URL being the URL prior to redirects. See step 2 of https://w3c.github.io/reporting/#generate-a-report That's what https://github.com/web-platform-tests/wpt/pull/56921 confirms is the existing behavior in browsers.

domfarolino:
> Window environments are handled separately in step 3."client".2 of https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer Therefore, we only use the creation URL for non-window global scopes (which are the workers).

Ah right, yes!

> Other specifications such as the reporting API rely on the creation URL being the URL prior to redirects.

I wouldn't phrase it like this. The Reporting API just relies on the creation URL, [period](https://github.com/w3c/reporting/pull/289). For Workers, that's the URL prior to redirects, and for Windows, that's the URL _after_ redirects since the [creation URL passed into "set up window environment settings object"](https://html.spec.whatwg.org/#shared-document-creation-infrastructure:set-up-a-window-environment-settings-object) is the navigation params's response's URL, which is the final URL.

> That's what https://github.com/web-platform-tests/wpt/pull/56921 confirms is the existing behavior in browsers.

I'm confused, what exactly does that PR confirm? The test expects that the the [`Report`'s url](https://w3c.github.io/reporting/#ref-for-dom-report-url) (which is just the Worker creation URL, from what I can tell) is the worker's current location href, which is the final URL after redirects. Further, https://wpt.fyi/results/content-security-policy/reporting-api/dedicated-worker-correct-url-in-report.html?label=pr_head&max-count=1&pr=56921 shows that Chrome passes the test, and other browsers have timeouts or mysterious "NOT RUN" statuses.

It seems really bad if the creation URL is the URL before redirects, and the reporting API exposes said URL through the `Report` object exposed in Workers, right? So it seems like the real bug is that the creation URL for Workers is never updated.

TimvdLippe:
All right, did some more digging and my initial assumption was wrong. I did before write the following assertion:

```js
assert_equals(reports[0].url, "{{location[server]}}/content-security-policy/reporting-api/support/dedicatedworker-report-url.sub.js");
```

This failed in Chrome, but I missed the part where it wasn't using the `redirect.py`:

```
assert_equals: expected "http://localhost:8000/content-security-policy/reporting-api/support/dedicatedworker-report-url.sub.js" but got "http://localhost:8000/content-security-policy/reporting-api/support/dedicatedworker-report-url.sub.js?location=%2Fcontent-security-policy%2Freporting-api%2Fsupport%2Fdedicatedworker-report-url.sub.js&test-name=%27+%28with+redirect%29%27&count=1"
```

So it does use the URL after redirecting, including the various other query parameters.

I have pushed an update to the test confirming that it uses the URL after redirect. So yes, you are correct in that HTML indeed missing the updating of the `creation_url`.

Note that Firefox fails the test since it doesn't support `ReportingObserver` yet. Safari I don't know since I don't have a Mac so I can't inspect why it fails. I suspect it's because Safari doesn't support `worker-src`.

domfarolino:
> Safari I don't know since I don't have a Mac so I can't inspect why it fails. I suspect it's because Safari doesn't support worker-src.

Maybe we can use a different fetch tactic, instead of nested Workers and worker-src? Can we just use the fetch API or something simpler? Not sure if that would help...

TimvdLippe:
Unfortunately we need to trigger a report and thus far I only know CSP can trigger that inside a worker. Maybe there is a different report we can generate, but for Servo we only integrate with CSP atm.

domfarolino:
I was just thinking maybe you could use `connect-src` or `default-src` and trigger the report via `fetch()` which violates the CSP, instead of relying on `worker-src` and `new Worker()` to make the network request, which you mentioned might cause issues in Safari.

----

So what is the next step here. Reading:

> So yes, you are correct in that HTML indeed missing the updating of the `creation_url`.

... I'm left to think that this PR should be closed in pursuit of the HTML one? Is that accurate?

TimvdLippe:
I initially used `connect-src` but couldn't make it work. I probably screwed something else up, since I have now implemented it and it runs as expected for Servo and Chrome. Hopefully now it also runs in Safari.

Yes, let's close this PR and reconvene on the HTML PR.

TimvdLippe:
FYI https://wpt.fyi/results/content-security-policy/reporting-api/dedicated-worker-correct-url-in-report.html?label=pr_head&max-count=1&pr=56921 still shows that Safari times out. I think the integration of CSP and workers in Safari are lacking, hence they fail these kinds of tests.

